PR #8, opened 2026-02-13
Add interactive model picker and simplify README

rishiskhare:
Adds a `claude-free` launcher script that lets you pick any NVIDIA NIM model from a searchable list before starting Claude Code. No more editing `.env` to switch models.

Also simplifies the README to make setup easier to follow — PM2 runs the proxy in the background by default, and fzf powers the model picker.


Alishahryar1:
With regards to the background feature, I opted against implementing it because if the user forgets about the process it's essentially orphaned and runs indefinitely, so, it's better to have it visible so the user can close it.

rishiskhare:
I found having to run a server in an external terminal very annoying, and the user can stop the server at any point. Would you like me to resolve the merge conflicts to add this in or are you against this change

PR #9, opened 2026-02-14
Fixing issue with kimi 2.5

Jeeltilva:


Alishahryar1:
Adding to this currently in main, thinking is only enabled when the client requests it otherwise it disables, the client smartly decides when to enable and disable thinking which should be preferred behavior, so please explain the purpose of this PR more elaborately.

PR #53, opened 2026-02-24
feat: Use NVIDIA NIM ASR for audio transcription

MauroDruwel:
## Summary
Added NVIDIA NIM as a second transcription option ( alongside local Whisper). This lets you transcribe voice notes using NVIDIA's cloud API instead of running Whisper locally.

## What changed

- **Transcription**: Now supports the two backends

  - Local Whisper: Free, runs on your GPU/CPU (existing)
  - NVIDIA NIM: Cloud API via Riva gRPC (new)

- **Supported models**: 8 NVIDIA NIM models added (Parakeet variants for different languages, Whisper Large V3)

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -144,7 +144,11 @@ class Settings(BaseSettings):
 
     @model_validator(mode="after")
     def check_nvidia_nim_api_key(self) -> Settings:
-        if self.whisper_device == "nvidia_nim" and not self.nvidia_nim_api_key.strip():
+        if (
+            self.voice_note_enabled
+            and self.whisper_device == "nvidia_nim"
+            and not self.nvidia_nim_api_key.strip()
+        ):
             raise ValueError(
                 "NVIDIA_NIM_API_KEY is required when WHISPER_DEVICE is 'nvidia_nim'. "
                 "Set it in your .env file."


MauroDruwel:
> Ok sounds good as long as it can be with uv sync --voice and doesn't require launching an extra server. You can do it.

Isn't it better to use something like `uv sync --voice_nim`? Some dependencies in `--voice` are over 1 GB and aren't needed for my part, they’re only required for local Whisper, so downloading them would be pointless.

MauroDruwel:
Something to note: the project uses a very recent Python version (3.14), and wheels for `grpcio-tools` are not yet available. As a result, on a fresh system you need to install the required build dependencies:

```bash
sudo apt install build-essential cmake pkg-config
```

Alishahryar1:
Is it ready to merge?

MauroDruwel:
@Alishahryar1 It's ready for review :wink:
Feel free to test around and play with it first, I haven't done much testing, but seems to work and be quite fast aswell

messaging/transcription.py
```diff
@@ -141,3 +170,68 @@ def _transcribe_local(file_path: Path, whisper_model: str, whisper_device: str)
...
+    # Read audio file
+    with open(file_path, "rb") as f:
+        data = f.read()
+
+    # Perform offline recognition
+    response = asr_service.offline_recognize(data, config)
+
+    # Extract text from response - cast to the expected result type
+    if response.results and response.results[0].alternatives:
+        transcript = response.results[0].alternatives[0].transcript
```
Alishahryar1: If transcript is not assigned due to some reason then logger will throw a name error

messaging/transcription.py
```diff
@@ -114,7 +134,16 @@ def transcribe_audio(
             f"Audio file too large ({size} bytes). Max {MAX_AUDIO_SIZE_BYTES} bytes."
         )
 
-    return _transcribe_local(file_path, whisper_model, whisper_device)
+    # Use settings values (from WHISPER_DEVICE and WHISPER_MODEL env vars)
+    # Function args are kept for backward compatibility but settings take precedence
+    settings = get_settings()
```
Alishahryar1: There is no need to read settings again here, args already use correct env vars at this function's call sites.

pyproject.toml
```diff
@@ -21,6 +21,11 @@ dependencies = [
 
 [project.optional-dependencies]
 voice = [
+    "grpcio>=1.78.0",
+    "grpcio-tools>=1.78.0",
```
Alishahryar1: Is grpc-tools needed shouldnt riva client come prebuilt with those?

.env.example
```diff
@@ -34,12 +34,17 @@ MESSAGING_RATE_LIMIT=1
...
+# Voice Note Transcription
 VOICE_NOTE_ENABLED=false
-# WHISPER_MODEL: Hugging Face ID or short name (tiny, base, small, medium, large-v2, large-v3, large-v3-turbo)
-WHISPER_MODEL="base"
-# WHISPER_DEVICE: "cpu" | "cuda"
-WHISPER_DEVICE="cpu"
+# WHISPER_DEVICE: "cpu" | "cuda" | "nvidia_nim"
+# - "cpu"/"cuda": Hugging Face transformers Whisper (offline, free; install with: uv sync --extra voice_local)
+# - "nvidia_nim": NVIDIA NIM Whisper via Riva gRPC (requires NVIDIA_NIM_API_KEY; install with: uv sync --extra voice)
+WHISPER_DEVICE="nvidia_nim"
```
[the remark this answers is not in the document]
MauroDruwel: Implemented in settings.py

config/settings.py
```diff
@@ -142,6 +142,15 @@ def validate_model_format(cls, v: str) -> str:
             )
         return v
 
+    @model_validator(mode="after")
+    def check_nvidia_nim_api_key(self) -> Settings:
+        if self.whisper_device == "nvidia_nim" and not self.nvidia_nim_api_key.strip():
```
Alishahryar1: Also add to the check that voice note is enabled

PR #73, opened 2026-03-05
Add token auth to all requests

xiaohang99:
I noticed that deploying the project publicly is risky since there's no auth, so I added token validation. It uses the ANTHROPIC_AUTH_TOKEN from the .env file. If the token is there, it'll check every request; if not, it just works like it did before without validation.

### DIFF
--- a/nvidia_nim_models.json
+++ b/nvidia_nim_models.json
@@ -638,6 +638,12 @@
             "owned_by": "nvidia"
         },
         {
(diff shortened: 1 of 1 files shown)


Alishahryar1:
This is a good feature, so I am assuming if I leave it empty then any auth_token works which is the current behaviour? It would be great if you could fix the failing checks.

Issue #67, opened 2026-03-02
dockerised?


jonnykking:
Any plans to dockerise Uvicorn so this could be a container run in a docker sandbox for an isolated claude?

boly38:
why is this issue closed? it's a good idea. 

--EDIT:
I just found & link some related PR with same answer from owner..

PR #79, opened 2026-03-08
Build and publish docker image

th-ch:
This PR builds and publishes a docker image to ghcr.io - it requires a "classic" Personal access token (stored as a secret with the name `GH_TOKEN`) and setting the package as public.


boly38:
related issue #67

PR #125, opened 2026-04-20
feat: add proxy support for httpx clients

arssing:
Add proxy support for providers based on [doc](https://www.python-httpx.org/advanced/proxies/):

  - Add per-provider proxy support (HTTP and SOCKS5) for all 4 providers: nvidia_nim, open_router, lmstudio, llamacpp
  - Each provider gets its own env var (NVIDIA_NIM_PROXY, OPENROUTER_PROXY, LMSTUDIO_PROXY, LLAMACPP_PROXY) for independent proxy configuration

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,12 @@ ENABLE_THINKING=true
 
 
 # Provider config
+# Per-provider proxy support: http and socks5, example: "http://username:password@host:port"
+NVIDIA_NIM_PROXY=""
+OPENROUTER_PROXY=""
+LMSTUDIO_PROXY=""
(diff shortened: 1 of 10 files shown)


Alishahryar1:
What is the benefit of this?

arssing:
1. Nvidia NIM not available in my region 
2. Someone may not have access to OpenRouter
3. If we do not have direct access to a subnet with deployed models (Llama and LM Studio), we can configure a proxy server on a host with models without having to configure a VPN. But this may be overthinking, so I can delete the code for Llama and LM Studio 😅

PR #151, opened 2026-04-24
feat: add Claude CLI compatibility via OpenAI chat completions bridge

Cupid-0x80h:
## Summary
Adds Claude CLI compatibility to the proxy by introducing an OpenAI-compatible `/v1/chat/completions` endpoint that bridges to the existing `/v1/messages` Anthropic pipeline.

## Changes
- Added `/v1/chat/completions` endpoint
- Implemented OpenAI → Anthropic message conversion
- Reused existing streaming pipeline (`create_message`)
- Ensures full Claude CLI compatibility

## Testing
- Tested with curl `/v1/chat/completions`
- Tested with Claude CLI
- Verified streaming responses and model routing (Qwen)

## Notes
This maintains backward compatibility with existing `/v1/messages` implementation and does not duplicate provider logic.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -225,3 +225,52 @@ async def stop_cli(request: Request, _auth=Depends(require_api_key)):
     count = await handler.stop_all_tasks()
     logger.info("STOP_CLI: source=handler cancelled_count={}", count)
     return {"status": "stopped", "cancelled_count": count}
+
+@router.post("/v1/chat/completions")
+async def chat_completions(
+    request: Request,
+    settings: Settings = Depends(get_settings),
+    _auth=Depends(require_api_key),
+):
+    """
+    OpenAI-compatible endpoint for Claude CLI.
+    Converts → Anthropic /v1/messages
+    """
+
+    body = await request.json()
+
+    messages = body.get("messages", [])
+    model = body.get("model")
+    stream = body.get("stream", False)
+
+    if not messages:
+        raise HTTPException(status_code=400, detail="messages cannot be empty")
+
+    #  Convert OpenAI → Anthropic format
+    anthropic_messages = []
+    system_prompt = None
+
+    for msg in messages:
+        if msg["role"] == "system":
+            system_prompt = msg["content"]
+        else:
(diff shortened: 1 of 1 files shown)


Alishahryar1:
Explain the reason for this PR, claude code doesn't need this route.

Cupid-0x80h:
While I was trying to use the project with Claude CLI and other OpenAI-compatible tools, I was getting errors because the system only exposed the /v1/messages  endpoint. When attempting to run requests through Claude CLI or OpenAI-style clients, it resulted in issues like 404 errors on /v1/chat/completions and request timeouts because those clients expect the OpenAI chat completions format

To fix this, I added a /v1/chat/completions route that acts as a compatibility layer. This endpoint converts OpenAI-style requests into the existing Anthropic /v1/messages format and reuses the existing provider and streaming pipeline instead of duplicating logic.

After implementing this change, the issue was resolved and both Claude CLI and OpenAI-compatible tools started working correctly with the proxy without modification on the client side.

On further research, I didn’t find an existing OpenAI-compatible bridge in this project, so this addition helps improve compatibility for users who want to use different LLM clients with the same backend. It also makes the system more flexible for multi-tool and multi-provider setups.

This should also help others who encounter the same issue where OpenAI-based tools fail due to missing /v1/chat/completions support.

Alishahryar1:
But this project is only for claude code which is not open au compatible

Cupid-0x80h:
While using this project with external LLM tooling, I encountered issues because only /v1/messages was exposed, causing OpenAI-style clients expecting /v1/chat/completions to fail with missing endpoint errors and timeouts.
To resolve this, I added an optional /v1/chat/completions compatibility layer that translates OpenAI requests into the existing /v1/messages pipeline without changing core Claude Code behavior.
This improves interoperability for external tools while keeping the project’s original Claude Code–focused design unchanged.

Alishahryar1:
I agree and you are right but, I don't play on supporting open ai compatible tools for now. But don't worry, I will revisit in the future. Closing PR for now.

PR #155, opened 2026-04-24
fix: assign cls._instance in __new__ to complete the GlobalRateLimiter singleton

### LINKED ISSUE #154: Bug: GlobalRateLimiter.__new__ never assigns cls._instance, breaking singleton when called directly
## Bug Report

**File:** `providers/rate_limit.py`

### Description

`GlobalRateLimiter` is designed as a singleton, but `__new__` never writes the newly-created instance back to `cls._instance`.

```python
def __new__(cls, *args, **kwargs):
    if cls._instance is not None:
        return cls._instance
    instance = super().__new__(cls)
    return instance          # ← cls._instance is never set!
```

Because `cls._instance` is never assigned in `__new__`, every direct instantiation with `GlobalRateLimiter(...)` creates a fresh object. The singleton *appears* to work only because `get_instance()` always writes `cls._instance` itself before returning — but any code that calls the constructor directly (including tests that call `GlobalRateLimiter(rate_limit=..., rate_window=...)`) gets a **different, non-shared** instance.

### Fix

Add the missing assignment:

```python
def __new__(cls, *args, **kwargs):
    if cls._instance is not None:
        return cls._instance
    instance = super().__new__(cls)
    cls._instance = instance   # ← store the new instance
    return instance
```

### Impact

- Any code that instantiates `GlobalRateLimiter` directly instead of going through `get_instance()` gets an unregistered object that will never be returned by subsequent calls to `get_instance()`, leading to multiple independent rate-limiter objects running concurrently and defeating the purpose of the singleton.

lingxiu58:
## Summary

Fixes a bug in `providers/rate_limit.py` where `GlobalRateLimiter.__new__` never stored the newly-created instance in `cls._instance`.

## Problem

```python
def __new__(cls, *args, **kwargs):
    if cls._instance is not None:
        return cls._instance
    instance = super().__new__(cls)
    return instance   # ← cls._instance is never set!
```

Every call to the constructor creates a fresh object because the guard `if cls._instance is not None` always sees `None`. The singleton only appears to work because `get_instance()` sets `cls._instance` before returning — but any code that calls the constructor directly (e.g. in tests or advanced usage) receives a **separate, unregistered** rate-limiter, defeating the purpose of the singleton and potentially allowing multiple concurrent rate-limiters with independent state.

## Fix

```python
def __new__(cls, *args, **kwargs):
    if cls._instance is not None:
        return cls._instance
    instance = super().__new__(cls)
    cls._instance = instance   # ← store the new instance
    return instance
```

One-line addition that makes `__new__` behave correctly as a singleton factory.

Closes #154

Signed-off-by: Cocoon-Break <86288566+lingxiu58@users.noreply.github.com>

### DIFF
--- a/providers/rate_limit.py
+++ b/providers/rate_limit.py
@@ -34,6 +34,7 @@ class GlobalRateLimiter:
         if cls._instance is not None:
             return cls._instance
         instance = super().__new__(cls)
+        cls._instance = instance
         return instance
 
     def __init__(


Alishahryar1:
This trades one issue for another I am planning a broader refactor anyway so closing.

PR #163, opened 2026-04-24
fix: assign cls._instance in GlobalRateLimiter.__new__ + add test

### LINKED ISSUE #154: Bug: GlobalRateLimiter.__new__ never assigns cls._instance, breaking singleton when called directly

fuleinist:
## Fix GlobalRateLimiter singleton broken by missing cls._instance assignment

**Problem:** GlobalRateLimiter.__new__ never assigns the newly created instance back to cls._instance, so direct instantiation (bypassing get_instance()) creates a separate, unregistered object. This causes multiple independent rate limiter instances to run concurrently, defeating the singleton's purpose.

**Fix:** Add `cls._instance = instance` before `return instance` in `__new__`.

**Also:** Added `test_direct_instantiation_returns_singleton` test to explicitly cover direct `GlobalRateLimiter(...)` calls, complementing the existing `test_singleton_pattern` test that only covered `get_instance()`.

Fixes #154

### DIFF
--- a/providers/rate_limit.py
+++ b/providers/rate_limit.py
@@ -34,6 +34,7 @@ class GlobalRateLimiter:
         if cls._instance is not None:
             return cls._instance
         instance = super().__new__(cls)
+        cls._instance = instance
         return instance
 
     def __init__(
--- a/tests/providers/test_provider_rate_limit.py
+++ b/tests/providers/test_provider_rate_limit.py
@@ -137,6 +137,23 @@ class TestProviderRateLimiter:
         assert limiter1 is limiter2
 
     @pytest.mark.asyncio
+    async def test_direct_instantiation_returns_singleton(self):
+        """
+        Direct calls to GlobalRateLimiter(...) must return the singleton.
+
+        Without the __new__ fix, each direct call creates a separate instance.
+        """
+        GlobalRateLimiter.reset_instance()
+        limiter1 = GlobalRateLimiter(rate_limit=10, rate_window=1)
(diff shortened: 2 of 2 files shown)


fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #180, opened 2026-04-25
Strengthen validation logic in NimSettings (pydantic model)

SEPURI-SAI-KRISHNA:
## Summary

This PR improves and tightens the Pydantic validation for `NimSettings` in `config/nim.py`, ensuring that all field inputs are robustly coerced, reject bad values early, and provide clear error messages for misconfigurations.

### What was changed
- Added type coercion and safety to fields allowing environment/CLI config values.
- Provided default values on empty strings and None where safe.
- Added detailed value checks for float fields and integer fields.
- Improved docstrings for field clarity.

---

### DIFF
--- a/config/nim.py
+++ b/config/nim.py
@@ -2,19 +2,18 @@
 
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 
-
 class NimSettings(BaseModel):
     """Fixed NVIDIA NIM settings (not configurable via env)."""
 
-    temperature: float = Field(1.0, ge=0.0)
-    top_p: float = Field(1.0, ge=0.0, le=1.0)
+    temperature: float = Field(1.0, ge=0.0, le=2.0, description="Sampling temperature, must be >=0 and <=2.")  # Added le=2.0 for safety
+    top_p: float = Field(1.0, ge=0.0, le=1.0, description="Nucleus sampling probability. [0,1]")
     top_k: int = -1
-    max_tokens: int = Field(81920, ge=1)
+    max_tokens: int = Field(81920, ge=1, description="Maximum number of tokens in output.")
     presence_penalty: float = Field(0.0, ge=-2.0, le=2.0)
     frequency_penalty: float = Field(0.0, ge=-2.0, le=2.0)
 
-    min_p: float = Field(0.0, ge=0.0, le=1.0)
-    repetition_penalty: float = Field(1.0, ge=0.0)
+    min_p: float = Field(0.0, ge=0.0, le=1.0, description="Minimum probability threshold [0,1].")
+    repetition_penalty: float = Field(1.0, ge=0.0, description="Penalty for repeated tokens. Must be >=0.")
 
     seed: int | None = None
     stop: str | None = None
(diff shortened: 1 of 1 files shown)


config/nim.py
```diff
@@ -2,49 +2,81 @@
 
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 
-
 class NimSettings(BaseModel):
     """Fixed NVIDIA NIM settings (not configurable via env)."""
 
-    temperature: float = Field(1.0, ge=0.0)
-    top_p: float = Field(1.0, ge=0.0, le=1.0)
+    temperature: float = Field(1.0, ge=0.0, le=2.0, description="Sampling temperature, must be >=0 and <=2.")  # Added le=2.0 for safety
```
Alishahryar1: `le=2.0` is a defensible cap for many APIs; worth double-checking against the NIM endpoint you use so you don’t reject valid server-side values.

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
+        if v is None or v == "":
+            return -1
+        int_v = int(v)
+        if int_v < -1:
             raise ValueError("top_k must be -1 or >= 0")
-        return v
+        return int_v
+
+    @field_validator("temperature", "top_p", "min_p", "presence_penalty",
+                     "frequency_penalty", "repetition_penalty", pre=True)
```
Alishahryar1: **Pydantic v2:** `field_validator` has no `pre=` keyword. This raises `TypeError: field_validator() got an unexpected keyword argument 'pre'` at class definition. Use `mode="before"` instead (same for the int validator on line 54).

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
+    @classmethod
+    def validate_float_fields(cls, v, field):
```
Alishahryar1: With `mode="before"`, the third parameter is `ValidationInfo`, not a Field. Rename accordingly and avoid treating it like a Pydantic `Field` (no `.default` / `.name` in the v1 sense).

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
+        # Allow string conversion if possible, otherwise fail
+        if v is None or v == "":
+            return field.default
```
Alishahryar1: `ValidationInfo` doesn’t provide `.default` here. Return explicit defaults per field (e.g. branch on `info.field_name`) or read defaults from the model’s field info in a v2-idiomatic way.

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
+        try:
+            val = float(v)
+        except Exception:
+            raise ValueError(f"{field.name} must be a float. Got {type(v).__name__}.")
```
Alishahryar1: Same: prefer `info.field_name` in the message instead of `field.name` once the parameter is `ValidationInfo`.

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
+        return val
+
+    @field_validator("max_tokens", "min_tokens", pre=True)
```
Alishahryar1: Same `pre=True` issue, use `mode="before"` for Pydantic 2.x.

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
+    @classmethod
+    def validate_int_fields(cls, v, field):
+        if v is None or v == "":
+            return field.default
```
Alishahryar1: Same as the float path: `field.default` / `field.name` need to be reworked for `ValidationInfo` + explicit defaults per field name.

config/nim.py
```diff
@@ -2,49 +2,81 @@
...
 
     @field_validator("stop", "chat_template", "request_id", mode="before")
     @classmethod
     def parse_optional_str(cls, v):
         if v == "":
             return None
-        return v
+        if v is not None and not isinstance(v, str):
+            return str(v)
+        return v
```
Alishahryar1: Please add a **trailing newline** at end of file (common style / POSIX).

Alishahryar1 (changes requested):
Thanks for tightening `NimSettings`—the direction (string coercion, empty string → defaults, clearer bounds) is useful. A few issues block merging against this repo’s **Pydantic 2.13+** stack; details are inline.

**Summary for the follow-up pass**
- Replace `pre=...` with `mode="before"` on `field_validator` (v2 removed `pre`).
- The validator receives `ValidationInfo` as the second argument, not a Field—avoid `field.default` / `field.name` without adapting to the v2 API (e.g. per-field defaults keyed by `info.field_name` or `model_fields`).
- Restore a **newline at EOF**.
- After fixes, **add tests** and run the project checks: `uv run ruff format`, `uv run ruff check`, `uv run ty check`, `uv run pytest` (CI enforces these).

SEPURI-SAI-KRISHNA:
@Alishahryar1 the issue mentioned above are resolved and i have locally run 
uv run ruff format
uv run ruff check
uv run ty check
uv run pytest

looks good to me.

PR #188, opened 2026-04-25
fix: Python 3.14 / Pydantic 2.13.3 compatibility

dyf102:
Implement a monkeypatch for typing._eval_type in core/compatibility.py to safely handle the missing prefer_fwd_module argument which caused TypeErrors during Pydantic model initialization on Python 3.14.0rc1.

- Add core/compatibility.py with the patch logic.
- Apply patch early in api, config, and providers packages.
- Update tests/conftest.py to force-override environment variables.
- Adjust import boundary tests to allow the compatibility module.
- Add regression test in tests/core/test_compatibility.py.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -1,0 +1,4 @@
+# TODO
+
+## Technical Debt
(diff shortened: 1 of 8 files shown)


Alishahryar1:
No need to do this just update your and install the correct python

PR #191, opened 2026-04-25
docs: fix quick-start uv installation instructions

TAGISWILD:
## Summary
The README quick-start path currently tells users to install `uv` with `pip install uv`.

## What I validated
I tried the smallest documented setup path locally and hit a real failure immediately:
- `python3.14 -m pip install uv`
- Result: `externally-managed-environment` (PEP 668) on Homebrew-managed Python

I also checked `pyproject.toml`, which requires Python `>=3.14`.

## Docs fix
- Replaced the `pip install uv` guidance with the official `uv` installer
- Added `uv self update`
- Added `uv python install 3.14`
- Noted why `pip install uv` is unreliable on this kind of macOS/Homebrew setup

## Scope
Docs only: `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,19 @@ A lightweight proxy that routes Claude Code's Anthropic API calls to **NVIDIA NI
 2. Install [Claude Code](https://github.com/anthropics/claude-code)
 
 ### Install `uv`
+
 ```bash
-# Install uv (required to run the project)
-pip install uv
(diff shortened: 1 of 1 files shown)


Alishahryar1:
Can you also include the command for power shell?

Alishahryar1:
If it's good enough to merge you can mark it ready

PR #197, opened 2026-04-26
feat: add toggle to disable provider throttling

Galzi1:
## Description
This PR introduces a new configuration option to completely bypass the internal provider-level throttling logic (proactive, reactive, and concurrency capping).

While the proxy's internal rate limiting is beneficial for most users, those with high-tier API limits or local provider setups (like LM Studio or Ollama) may prefer to disable these checks to reduce latency and allow the upstream provider to handle request volume entirely.

### Key Changes
- **Configuration**: Added `ENABLE_PROVIDER_THROTTLING` environment variable (defaults to `True`).
- **Rate Limiter**: Updated `GlobalRateLimiter` to respect the `enabled` flag. When disabled:
    - `wait_if_blocked()` returns immediately.
    - `concurrency_slot()` yields without acquiring a semaphore.
    - `execute_with_retry()` executes the request directly without internal retries on 429s (fail-fast behavior).
- **Documentation**: Added `docs/throttling.md` explaining how to use the toggle.
- **Verification**: Added unit tests in `tests/providers/test_rate_limit_toggle.py` to ensure the bypass logic works as expected.

### How to use
Set the following in your `.env` file:
```env
ENABLE_PROVIDER_THROTTLING=False
```

### Testing
- Created new unit tests verifying that all limiter modes (wait, concurrency, and retry) are bypassed when the toggle is off.
- Verified that default behavior remains unchanged (throttling active by default).
- Confirmed that settings are correctly propagated through the provider registry to individual transport instances.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -1,5 +1,7 @@
 """Centralized configuration using Pydantic Settings."""
 
+from __future__ import annotations
+
 import os
 from collections.abc import Mapping
 from functools import lru_cache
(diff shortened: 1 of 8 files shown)


Alishahryar1:
Why not just raise to very high limits?

PR #259, opened 2026-04-29
feat(nim): auto-prefix vendor namespace for short NIM model names

Moiz-Ali-Moomin:
## Summary
- Users can now set `MODEL_OPUS=nvidia_nim/qwen3-coder-480b-a35b-instruct` or `nvidia_nim/gemma-3-27b-it` without knowing each model's NIM vendor path. The request builder rewrites short names to their vendor form (`qwen/`, `deepseek-ai/`, `mistralai/`, `z-ai/`, `google/`, `meta/`, `nvidia/`) before the upstream call.
- Names that already contain `/` pass through unchanged, so existing configs like `nvidia_nim/z-ai/glm-4.7` keep working with no behavior change.
- Unknown families also pass through unchanged, so adding a new vendor doesn't require a code change when the user configures the full path.

## Motivation
NVIDIA NIM exposes most chat models under a vendor namespace (`google/gemma-3-27b-it`, `qwen/qwen3-coder-480b-a35b-instruct`, etc.). Without the prefix, the upstream returns a 404 and the failure surfaces as an opaque provider error. New users have to read NVIDIA's catalog page just to fill in the right string for `MODEL_OPUS` / `MODEL_SONNET` / `MODEL_HAIKU`. This PR removes that friction for the most common families.

## Implementation
- New `map_nim_model()` helper and `_NIM_VENDOR_PREFIXES` table in [providers/nvidia_nim/request.py](providers/nvidia_nim/request.py).
- Called once inside `build_request_body()` right after `build_base_request_body()` returns, so the rewrite is contained to NIM and doesn't touch the shared OpenAI base or any other provider.
- Data-driven prefix table makes adding a new vendor a one-line change.

## Test plan
- [x] `uv run ruff format` — 225 files already formatted
- [x] `uv run ruff check` — all checks passed
- [x] `uv run ty check` — all checks passed
- [x] `uv run pytest tests/providers/` — 352 passed (18 new tests in `TestMapNimModel`)
- [x] New parametrized tests cover: each vendor prefix, already-prefixed pass-through, unknown prefix, `None` / `""` input, and end-to-end rewrite via `build_request_body`.

## Backwards compatibility
None broken. Configs that already use the full vendor path (e.g. `nvidia_nim/z-ai/glm-4.7`) hit the early `"/" in model` branch and are returned unchanged.

## Out of scope
- No change to non-NIM providers.
- No change to request body shape, headers, or streaming behavior.
- No change to model listing endpoints — this is purely an outbound rewrite.

### DIFF
--- a/providers/nvidia_nim/request.py
+++ b/providers/nvidia_nim/request.py
@@ -15,6 +15,34 @@ from core.anthropic import (
 from core.anthropic.conversion import OpenAIConversionError
 from providers.exceptions import InvalidRequestError
 
+# Map a short model family prefix to its NIM vendor namespace. Users can set
+# MODEL_OPUS=nvidia_nim/qwen3-coder-480b-a35b-instruct (no vendor prefix) and
+# the request builder will rewrite it to qwen/qwen3-coder-480b-a35b-instruct
+# before the upstream call. Names that already contain "/" are passed through.
+_NIM_VENDOR_PREFIXES: tuple[tuple[str, str], ...] = (
+    ("qwen", "qwen"),
+    ("deepseek", "deepseek-ai"),
+    ("devstral", "mistralai"),
+    ("glm", "z-ai"),
+    ("gemma", "google"),
+    ("llama", "meta"),
+    ("nemotron", "nvidia"),
+)
+
+
+def map_nim_model(model: str | None) -> str | None:
+    """Rewrite a short NIM model name to its vendor-prefixed form.
+
+    Returns ``model`` unchanged when it is empty, already vendor-prefixed
+    (contains "/"), or has no known vendor mapping.
+    """
+    if not model or "/" in model:
+        return model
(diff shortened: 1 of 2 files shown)


Alishahryar1:
This is brittle and bad practice

Moiz-Ali-Moomin:
can you elaborate please ?

Moiz-Ali-Moomin:
Hey, thanks for the feedback on #259. You were right — the hardcoded prefix table was brittle.

I've addressed it in #294 with a proper fix: instead of a static vendor map, the provider now calls `GET /v1/models` on the NIM endpoint to fetch the live model catalog, and does a suffix match to resolve short names. The mapping comes directly from NIM itself so it never goes stale — new vendors NVIDIA adds work automatically with zero code changes.

The fetch is cached per process (`lru_cache` on `base_url + api_key`) so it's a single HTTP call, and it degrades gracefully if the endpoint is unavailable.

PR #282, opened 2026-04-30
feat(ollama): OLLAMA_API_KEY support + ollama-only .env cleanup

peterlodri-sec:
## Summary
- `provider_catalog.py`: switch ollama from hardcoded `static_credential="ollama"` to `credential_attr="ollama_api_key"` — enables LiteLLM master key as Bearer auth when routing through the container stack
- `settings.py`: add `OLLAMA_API_KEY` field (default `"ollama"` for direct ollama, set to `LITELLM_MASTER_KEY` for stack mode)
- `.env.example`: strip to ollama-only; two-mode comments (Option A: stack :4000 / Option B: self-hosted :11434)

## Why
Running ollama behind a LiteLLM proxy (for prompt caching) requires authenticating with the proxy's master key. Previously the API key was hardcoded to `"ollama"` in the provider descriptor.

## Test plan
- [ ] `uv run ruff format` passes
- [ ] `uv run ruff check` passes
- [ ] `uv run ty check` passes
- [ ] Start server with `OLLAMA_BASE_URL=http://localhost:4000 OLLAMA_API_KEY=sk-test` — confirm request is sent with `Authorization: Bearer sk-test`
- [ ] Start server with no `OLLAMA_API_KEY` — confirm default `"ollama"` auth still works

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,111 +1,52 @@
-# NVIDIA NIM Config
-NVIDIA_NIM_API_KEY=""
-
-
-# OpenRouter Config
-OPENROUTER_API_KEY=""
-
-
-# DeepSeek Config (uses native Anthropic Messages at api.deepseek.com/anthropic)
-DEEPSEEK_API_KEY=""
-
-
-# LM Studio Config (local provider, no API key required)
-LM_STUDIO_BASE_URL="http://localhost:1234/v1"
-
-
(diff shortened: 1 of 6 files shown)


peterlodri-sec:
Closing — changes will be maintained as a local patch instead of upstream PR.

PR #288, opened 2026-04-30
feat: improve type safety in Settings validators in config/settings.py

rayrishu19-wq:


Alishahryar1:
I am not looking for these sorts of PRs

PR #289, opened 2026-04-30
feat: improve Ollama base URL validation error message in config/sett…

rayrishu19-wq:
…ings.py


Alishahryar1:
Same thing repeated

PR #318, opened 2026-05-02
feat: MAX_MESSAGES + CONTEXT_MAX_TOKENS (tool-aware, seam-safe) + NIM_PARALLEL_TOOL_CALLS + SSE fix + Claude 4 IDs + tool schema hint

zhapostolski:
### Context truncation (tool-aware, seam-safe)
- **MAX_MESSAGES** — trim conversation to last N messages before forwarding
- **CONTEXT_MAX_TOKENS** — token-budget trimming counts tool schemas (~200 tok × N tools) in the budget, not just messages
- **Seam sanitizer** — after trimming, drops leading messages until the conversation starts at a clean user turn; prevents 400 Bad Request when aggressive cuts orphan a `tool_use` block with no matching `tool_result`
- **`min_messages` floor** — prevents over-trimming to 0 messages on tool-heavy conversations
- **Bounded seam-walker fallback** *(new)* — when the unbounded walker would push results below `min_messages` (which happens routinely in Claude Code where every user turn is a tool_result), fall back to a 4-drop max bounded variant. Observed in production as `CONTEXT: trimmed 35 → 0 messages` mid-session — fix prevents catastrophic context wipeout.
- **Consecutive assistant dedup** — removes back-to-back assistant messages at the trim seam that NIM rejects

### Per-provider context budgets *(new)*
Tune NIM and OpenRouter independently:
- `NIM_CONTEXT_MAX_TOKENS`, `NIM_CONTEXT_MIN_MESSAGES`, `NIM_MAX_MESSAGES`
- `OPENROUTER_CONTEXT_MAX_TOKENS`, `OPENROUTER_CONTEXT_MIN_MESSAGES`, `OPENROUTER_MAX_MESSAGES`

Set per-provider value (>0) to override the global default for that provider only. Lets you keep NIM tight for free-tier latency (5-min stream cap, 256k context) while letting OR Qwen models use their full 1M-context window. `services.py` resolves routing before context-trim so the right provider's budget applies.

### Heuristic parser: `<function=NAME>` without leading bullet *(new)*
Some models (nemotron-super under heavy tool-loaded sessions, others) emit OpenAI-XML-style tool calls without the `●` bullet. The existing parser's `_FUNC_START_PATTERN` required the bullet, so those blocks leaked as raw XML to the client. Added `_FUNCTION_BLOCK_PATTERN` to catch the bullet-less variant in one regex; `<parameter=KEY>VAL</parameter>` extraction unchanged.

### OpenRouter improvements *(new)*
- **`OPENROUTER_INCLUDE_PAID_MODELS=true`** — opt-in to surface paid OR models in `/model` menu (default: free-only filter preserved). Lets paid Qwen models (e.g. `qwen3-235b-a22b-2507`, `qwen3-coder`, `qwen3-coder-flash`) appear after credit top-up.
- **`OPENROUTER_ENABLE_THINKING`** — provider-scoped thinking flag.
- **Auto-detect `-thinking` model variants** by name on OR routes (`qwen3-235b-a22b-thinking-2507` → reasoning mode auto-on without user config).

### Provider / model fixes
- **NIM_PARALLEL_TOOL_CALLS** — env flag to serialize tool calls for models that return 400 on parallel calls
- **Claude 4 short-form IDs** — `sonnet-4-6`, `opus-4-7`, `haiku-4-5` resolve correctly
- **Tool schema hint** — injected into system prompt so models without native tool-schema context still see argument structure

### Performance & stability
- **`model_copy` instead of `deepcopy`** — O(1) vs O(n_tools) per trimmed request
- **Async log enqueue** — removes startup truncation; all first-line logs now preserved
- **Fix: tokens.py syntax error** — `except TypeError, ValueError:` was Python 2 syntax; fixed

## Environment variables
| Variable | Default | Description |
|---|---|---|
| `MAX_MESSAGES` | 0 | Max messages sent to provider (0 = unlimited) |
| `CONTEXT_MAX_TOKENS` | 0 | Token budget including tool schemas (0 = unlimited) |
| `CONTEXT_MIN_MESSAGES` | 20 | Floor for seam-walker fallback |
| `NIM_MAX_MESSAGES` | 0 | NIM-specific override (0 = inherit global) |
| `NIM_CONTEXT_MAX_TOKENS` | 0 | NIM-specific override |
| `NIM_CONTEXT_MIN_MESSAGES` | 0 | NIM-specific override |
| `OPENROUTER_MAX_MESSAGES` | 0 | OR-specific override |
| `OPENROUTER_CONTEXT_MAX_TOKENS` | 0 | OR-specific override |
| `OPENROUTER_CONTEXT_MIN_MESSAGES` | 0 | OR-specific override |
| `OPENROUTER_INCLUDE_PAID_MODELS` | false | Surface paid OR models in /model picker |
| `OPENROUTER_ENABLE_THINKING` | false | Reasoning mode for OR routes |
| `NIM_PARALLEL_TOOL_CALLS` | true | Set false to serialize tool calls |

## Test plan
- [ ] Long session (>120k tokens) trims cleanly, no `trimmed N → 0` events
- [ ] Tool-heavy session with 91+ tools stays within token budget
- [ ] `<function=NAME>...</function>` without bullet emits `tool_use` block (not raw text)
- [ ] `OPENROUTER_INCLUDE_PAID_MODELS=true` lists paid Qwen models in `/v1/models`
- [ ] Default behavior (`OPENROUTER_INCLUDE_PAID_MODELS` unset) lists free models only — backward compatible
- [ ] `qwen3-235b-a22b-thinking-2507` auto-enables thinking on OR routes
- [ ] `NIM_CONTEXT_MAX_TOKENS=120000` + `OPENROUTER_CONTEXT_MAX_TOKENS=800000` apply per-provider when routing
- [ ] `sonnet-4-6` / `opus-4-7` resolve to correct NIM endpoints

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,6 +25,7 @@ OLLAMA_BASE_URL="http://localhost:11434"
 # All Claude model requests are mapped to these models, plain model is fallback
 # Format: provider_type/model/name
 # Valid providers: "nvidia_nim" | "open_router" | "deepseek" | "lmstudio" | "llamacpp" | "ollama"
+# MODEL="open_router/tencent/hy3-preview:free"
 MODEL_OPUS=
 MODEL_SONNET=
 MODEL_HAIKU=
--- a/INICIAR_CLAUDE_GRATIS.bat
+++ b/INICIAR_CLAUDE_GRATIS.bat
@@ -1,0 +1,37 @@
+@echo off
+title Claude Code Gratis (NIM Proxy)
+color 0A
+
+echo.
+echo  ==========================================
+echo   CLAUDE CODE GRATIS - NVIDIA NIM
+echo  ==========================================
+echo.
+
+REM Verificar si el proxy ya esta corriendo
+curl -s http://localhost:3000/health >nul 2>&1
+if %errorlevel% equ 0 (
+    echo  [OK] Proxy ya esta corriendo
+    goto :iniciar_claude
+)
+
+echo  Iniciando servidor proxy...
+start "NIM Proxy" /min cmd /c "cd /d C:\Users\titan\free-claude-code && uv run python server.py"
+
+echo  Esperando que el proxy arranque...
+:esperar
+timeout /t 2 /nobreak >nul
+curl -s http://localhost:3000/health >nul 2>&1
+if %errorlevel% neq 0 goto :esperar
+
+:iniciar_claude
+echo  [OK] Proxy activo en localhost:3000
+echo  Modelo: Llama 3.1 70B (NVIDIA NIM - GRATIS)
+echo.
+echo  Iniciando Claude Code...
+echo.
+
+set ANTHROPIC_BASE_URL=http://localhost:3000
+set ANTHROPIC_AUTH_TOKEN=freecc
+
+claude
--- a/INICIAR_CLAUDE_GRATIS.bat
+++ b/INICIAR_CLAUDE_GRATIS.bat
@@ -26,7 +26,7 @@ if %errorlevel% neq 0 goto :esperar
 
 :iniciar_claude
 echo  [OK] Proxy activo en localhost:3000
-echo  Modelo: Llama 3.1 70B (NVIDIA NIM - GRATIS)
+echo  Modelo: Llama 3.1 8B (NVIDIA NIM - GRATIS)
 echo.
 echo  Iniciando Claude Code...
 echo.
--- a/api/model_router.py
+++ b/api/model_router.py
@@ -50,7 +50,9 @@ class ModelRouter:
             thinking_enabled = (
                 force_thinking_enabled
                 if force_thinking_enabled is not None
-                else self._settings.resolve_thinking(direct_provider_model)
+                else self._settings.resolve_thinking(
+                    direct_provider_model, provider_id=direct_provider_id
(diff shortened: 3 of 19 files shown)


zhapostolski:
**Update: Added minimum message floor to prevent over-trimming**

Root-cause identified for sessions stopping mid-task: with `CONTEXT_MAX_TOKENS=55000` and thinking enabled (2–10k tokens per assistant turn), 62 tool schemas (~12k tokens), and a large system prompt (~15k), the token-trimming loop would reduce a 580-message context to **5–7 messages**. The model sees almost no history, writes a brief summary, and emits `finish_reason=stop` → agent halts.

Latest commit adds:
- `min_messages` floor (default **20**) to `ContextManager` — the trimming loop stops at `max(4, min_messages)` even if still over budget. This ensures the model always has meaningful context rather than hallucinating an end state.
- `CONTEXT_MIN_MESSAGES` env var wired through `Settings` → `get_context_manager()`
- Raised `CONTEXT_MAX_TOKENS` from 55 000 to **90 000** (closer to NIM's 128k limit, leaving ~38k for output + thinking)

Together with the existing seam-sanitizer and tool-schema hint, this should resolve premature session termination.

PR #360, opened 2026-05-05
fix: accept betas body field from Claude Code clients

aaravsw3:
Claude Code newer versions send a `betas` list in the request body (e.g. "interleaved-thinking-2025-05-14"). This landed in `__pydantic_extra__` and triggered `_openai_reject_native_only_top_level_fields`, surfacing as "Invalid request sent to provider." for all OpenAI Chat upstreams (kimi, nvidia_nim).

Declare `betas` on both `MessagesRequest` and `TokenCountRequest` with `exclude=True` so it is accepted from clients but never forwarded to any provider body.


Alishahryar1 (approved):
LGTM

PR #379, opened 2026-05-07
fix: convert Anthropic image blocks to OpenAI image_url format

### LINKED ISSUE #362: Can't respond/see pasted image in VScode
When I ALT + V (which is how you paste an image in Claude code) and send the image I get this error:

Image****

How can I send images and get a normal response? (and of course that he'll be able to see it)

AashishKumar-3002:
Previously the OpenAI chat converter raised an OpenAIConversionError for any user message image block, causing a 400 for providers like Kimi k2 that support vision.

Fix by adding _anthropic_image_to_openai() which maps:
- base64 source -> data:<media_type>;base64,<data> URL
- url source    -> the URL directly

Both _convert_user_message and _convert_user_message_with_injection now emit image_url content parts instead of raising. Text-only messages still use a plain string content field (backward compatible); mixed text+image messages use a content array as required by OpenAI.

Fixes #362


AashishKumar-3002:
@Alishahryar1 Hi! This PR fixes the image pasting issue reported in #362 — 
users can't send images via Alt+V/Pasted when using OpenAI-compatible providers 
(Kimi k2, OpenRouter, etc.). Happy to adjust anything if needed!

PR #380, opened 2026-05-07
Backported to use python3.13

dercol1:
* Backported (AI helped) to python3.13 as is currently installed in normal linux distro


Alishahryar1:
I want to stick to 3.14

PR #388, opened 2026-05-08
Fix startup crash on read-only filesystems by falling back to stderr logging

AashishKumar-3002:
## Summary
- Prevent startup/import failure when the configured log file path is not writable (for example, serverless read-only filesystems such as Vercel).
- Keep logging available by falling back to a stderr sink when file sink initialization fails.
- Add regression coverage for the read-only filesystem path.

## Root Cause
`configure_logging()` attempted to truncate/create `server.log` during app startup. On read-only filesystems this raised `OSError: [Errno 30] Read-only file system`, causing module import and process startup to fail.

## Changes
- Wrap file-log setup in `try/except OSError` in `config/logging_config.py`.
- On failure, initialize loguru with `sys.stderr` sink and emit a warning.
- Add test `test_configure_logging_falls_back_when_log_file_unwritable` in `tests/config/test_logging_config.py`.

## Validation
- `uv run pytest tests/config/test_logging_config.py`
- `uv run ruff format`
- `uv run ruff check`
- `uv run ty check`
- `uv run pytest`

All checks passed.

### DIFF
--- a/config/logging_config.py
+++ b/config/logging_config.py
@@ -9,6 +9,7 @@ included at top level for easy grep/filter.
 import json
 import logging
 import re
+import sys
 from pathlib import Path
 
 from loguru import logger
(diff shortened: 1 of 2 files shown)


Alishahryar1:
I don't want to support this

PR #399, opened 2026-05-09
feat(model-cache): persist provider model lists + periodic refresh

zhapostolski:
## Summary

Adds disk persistence and periodic background refresh for the provider model-discovery cache so that:

- **Cold proxy start serves `/v1/models` instantly from disk** while a fresh upstream discovery runs in the background. No more 1–2s lag before Claude Code's `/model` picker populates.
- **Long-running daemons stay current** — every 6h (configurable via `FCC_MODEL_REFRESH_HOURS`) every eligible provider is re-queried, picking up newly released models without a restart.
- **Cache survives restarts**, so the picker has the full known model list immediately even before any upstream call completes.
- **Manual refresh** via `POST /v1/models/refresh` for tooling that wants to force-refresh.

## Files changed

- `providers/model_cache_persistence.py` *(new)* — atomic JSON load/save helpers. Cache file at `~/.cache/free-claude-code/models.json` (override via `FCC_MODEL_CACHE_PATH`).
- `providers/registry.py` — hydrate-from-disk in `__init__`, persist after `_refresh_model_ids`, plus `start_periodic_model_list_refresh`, `force_refresh_all`, `clear_disk_cache`. Periodic task is cancelled in `cleanup`.
- `api/runtime.py` — wires the periodic refresh interval from env on startup. Default 6h; set `FCC_MODEL_REFRESH_HOURS=0` to disable.
- `api/routes.py` — `POST /v1/models/refresh` endpoint, returns `{status, providers, total}`.

Net: **+274 lines, no deletions**. No existing behavior changes when env vars are unset.

## Configuration

| Env var | Default | Purpose |
|---|---|---|
| `FCC_MODEL_REFRESH_HOURS` | `6` | Background refresh interval in hours. `0` disables. |
| `FCC_MODEL_CACHE_PATH` | `~/.cache/free-claude-code/models.json` | Override cache file path. |

## Test plan

- [x] Python syntax check passes for all four touched files.
- [x] Cold start with no cache file → behaves exactly like before (live discovery runs, then writes cache).
- [x] Restart with existing cache file → log shows `Provider model discovery hydrated from disk: providers=N saved_at=...`, in-memory cache populated, then existing `start_model_list_refresh` finds cache warm and skips redundant upstream calls.
- [x] `POST /v1/models/refresh` → returns 200 with `{"status":"ok","providers":{"nvidia_nim":131,"open_router":273},"total":404}` on a configured proxy.
- [x] `cleanup()` cancels the periodic task without leaking.
- [x] Cache file written atomically (tmp file → rename).
- [ ] Reviewer: please verify behavior on a config with only one provider configured (the loop should still log success, no surprise warnings).

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -236,6 +236,32 @@ async def list_models(
     return _build_models_list_response(settings, provider_registry)
 
 
+@router.post("/v1/models/refresh")
+async def refresh_models(
+    request: Request,
+    settings: Settings = Depends(get_settings),
+    _auth=Depends(require_api_key),
+):
+    """Force a synchronous refresh of the provider model-discovery cache."""
+    registry = getattr(request.app.state, "provider_registry", None)
+    if not isinstance(registry, ProviderRegistry):
+        raise HTTPException(
+            status_code=503, detail="Provider registry not initialized"
+        )
+    try:
+        counts = await registry.force_refresh_all(settings)
+    except Exception as exc:
+        logger.warning(
+            "Manual model refresh failed: exc_type={}", type(exc).__name__
+        )
+        raise HTTPException(status_code=502, detail="Refresh failed") from exc
+    return {
+        "status": "ok",
+        "providers": counts,
+        "total": sum(counts.values()),
+    }
+
+
 @router.post("/stop")
 async def stop_cli(request: Request, _auth=Depends(require_api_key)):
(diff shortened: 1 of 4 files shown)


Alishahryar1:
Claude already has model cache, internally

PR #408, opened 2026-05-10
feat(nim-resilience-1): transient retries + stream silence watchdog

zhapostolski:
## Summary

Phase 1 of a 4-phase plan to make NIM sessions survive transient upstream failures. This PR ships **transport-level resilience** — the things that today silently break long agentic sessions.

## Real-world failures this addresses

Observed in production logs from both fcc and opencode against NIM:

| Failure mode | Frequency | Today's behaviour | After this PR |
|---|---|---|---|
| HTTP 504 mid-stream (5-min cap) | Common on long thinking turns | Session dies, no retry | Retried with backoff if pre-stream; killed cleanly within 90s if mid-stream |
| HTTP 502/503 transient gateway | Routine | Propagates to user | Retried transparently |
| 404 ""Function not found"" | NIM serverless cold-start on big models | Hard fail | Retried after backoff |
| ``RemoteProtocolError`` mid-stream | Connection drop | Propagates | Retried if pre-stream |
| ``ReadTimeout`` (8.5h hang observed) | NIM keeps connection open with no bytes | httpx default ~5min | Watchdog cuts at 90s |
| 429 rate limit | Routine | Already retried | Same |

### `providers/transient_errors.py` (new)

Pure classifier: ""is this exception retryable?"". Centralises every "yes" decision. Includes a 404-with-body-marker check for NIM cold-start 404s.

### `providers/rate_limit.py` (extended)

`execute_with_retry` now retries every transient error (not just 429). Same exponential backoff. New env vars:

- `NIM_RETRY_MAX_ATTEMPTS` (default 4)
- `NIM_RETRY_BASE_DELAY_S` (2.0)
- `NIM_RETRY_MAX_DELAY_S` (60.0)
- `NIM_RETRY_JITTER_S` (1.0)
- `NIM_RETRY_ON_5XX` (default `408,429,502,503,504,520,522,524`)

Structured telemetry: every retry decision logs as `NIM_RETRY: cause=… attempt=N/M backoff=Xs elapsed=Ys`.

### `core/stream_watchdog.py` (new)

Async iterator wrapper that raises `StreamSilentError` if no chunk arrives within `silence_timeout_s` (default 90s, env `NIM_STREAM_SILENCE_TIMEOUT_S`). Closes the upstream iterator on trip. Wired into `providers/openai_compat.py` so every NIM stream gets watchdog protection.

### `docs/nim-resilience.md` (new)

Full design doc covering Phases 1-4. Phase 1 is in scope; Phases 2-4 (time-safe compaction, session memory, resume tokens) will follow as separate PRs.

## Test plan

- [x] **46 new unit tests** covering every transient error class, retry attempt path, env-override behavior, watchdog silence detection, and graceful close.
- [x] **Full upstream test suite passes**: 1282 tests pass (was 1236 before; +46 new, 0 regressions).
- [x] **Synthetic NIM 504 retry test**: simulated `httpx.HTTPStatusError(504)` → retries 3× with backoff → eventually succeeds.
- [x] **NIM cold-start 404 retry test**: 404 with NIM-specific body marker → retries; generic 404 → propagates immediately.
- [x] **Watchdog test**: 0.05s timeout + 1s sleep → `StreamSilentError` raised, upstream `aclose()` called.
- [ ] Reviewer: verify against a real NIM session by inducing a 502 (e.g. block NIM endpoint briefly via firewall) and observing `NIM_RETRY:` log lines + transparent recovery.

## Out of scope (Phases 2-4)

- Time-safe context compaction (Phase 2) — keep NIM input under 80k to avoid the 5-min cap by construction
- Hierarchical session memory (Phase 3) — SQLite-backed checkpoints across HTTP requests
- Resume tokens (Phase 4) — synthesize ""continuing"" responses on forced abort

Each phase will land as its own PR. See `docs/nim-resilience.md` for the full architecture.

### DIFF
--- a/core/stream_watchdog.py
+++ b/core/stream_watchdog.py
@@ -36,25 +36,78 @@ T = TypeVar("T")
 _DEFAULT_SILENCE_TIMEOUT_S = 90.0
 _SILENCE_TIMEOUT_ENV = "NIM_STREAM_SILENCE_TIMEOUT_S"
 
+# Tier-aware silence timeouts. Big models genuinely think for longer
+# without emitting tokens, so they get more rope before the watchdog
+# trips. Tiers are matched against the model id via substring patterns
+# (largest first). Frontier models (Claude/GPT-5/etc.) bypass entirely.
+_TIER_SILENCE_TIMEOUTS_S = {
+    "xxl": 180.0,    # mistral-large-3-675b, future Qwen3-Max
+    "xl": 150.0,     # qwen3-coder-480b, llama-3.1-405b, ring-1t
+    "large": 120.0,  # nemotron-ultra-253b, deepseek-v4-pro
+    "medium": 90.0,  # nemotron-super-120b, gpt-oss-120b
+    "small": 60.0,   # qwen3-next-80b, llama-70b
+    "tiny": 45.0,    # nemotron-nano-30b, gpt-oss-20b
+}
 
-def silence_timeout_s() -> float:
-    """Read the configured silence timeout from env, with a safe default."""
+# Same patterns/order as agent_governor.tiers.TIER_PATTERNS — kept here
+# duplicated to keep core/ free of api/ imports per the boundary check.
+_TIER_PATTERNS = (
+    ("xxl", ("675b", "qwen3-max", "mistral-large-3", "deepseek-r2")),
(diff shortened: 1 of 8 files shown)


Alishahryar1:
I already added retries for 5xx, you may need to rebase. I like this direction. 

I do not like the watchdog there is some other bug you due which it's doing this.

After rebasing see what still needs retries.

PR #465, opened 2026-05-17
Take into account Retry-After for provider limits & Fix the launch of Claude CLI from Telegram on Windows

### LINKED ISSUE #3: Error at work
Sometimes we get an error when working 

Provider rate limit reached. Please retry shortly. (request_id=req_2078e3d07718)

# Model Routing
MODEL=nvidia_nim/nvidia/nemotron-3-super-120b-a12b
MODEL_OPUS=nvidia_nim/qwen/qwen3-coder-480b-a35b-instruct
MODEL_SONNET=nvidia_nim/stepfun-ai/step-3.5-flash
MODEL_HAIKU=open_router/deepseek/deepseek-v4-flash:free

You can also see this in the proxy logs:

INFO:     127.0.0.1:57822 - "HEAD / HTTP/1.1" 204 No Content

We need to fix it all and make it work without these bugs.

labtgbot:
Fixes labtgbot/free-claude-code#3

## How to reproduce

- The provider returns HTTP 429 with a `Retry-After` header, but the local retry-loop only uses its own exponential backoff. With short retries, this can exhaust attempts before the provider will accept the request again, and the user sees `Provider rate limit reached.` Please retry shortly.`
- Uvicorn access log prints utility probe requests like `HEAD / HTTP/1.1" 204 No Content`, although the application already logs its own structured trace/error logs.

## Chto izmeneno

- `GlobalRateLimiter.execute_with_retry()` now reads `Retry-After` from `httpx.HTTPStatusError` and OpenAI SDK errors, supports delta-seconds and HTTP-date, and does not retry the request before the provider's window.
- `retry_after_s` has been added to the retry trace so that the logs show when the delay came from the upstream.
- Running via `fcc-server` and direct `server.py` disables uvicorn access log so that compatible probe requests don't clutter the proxy logs.

## Verification

- `uv run ruff format`
- `uv run ruff check`
- `uv run ty check`
- `uv run pytest` (1323 passed)

Fixes labtgbot/free-claude-code#5

## What has been changed

[Files Changed]
- `cli/session.py `: added resolution `CLAUDE_CLI_BIN` via `shutil.which()` for Windows, where npm publishes `claude.cmd'; the Claude CLI startup error now receives a clear message with an installation/configuration prompt.
- `tests/cli/test_cli.py `: added regression tests for Windows npm shim and for the missing command without the raw `[WinError 2]`.

[Logic Altered]
- Telegram/Discord messaging path via `CLISession.start_task()` now launches the found Windows shim instead of the raw `claude`.
- If the subprocess returns a `FileNotFoundError` anyway, the user will see an actionable error: install Claude Code or specify `CLAUDE_CLI_BIN` in the full path.
- The temporary `.gitkeep` has been removed from the initial commit; the net PR diff contains only code and tests.

[Verification Method]
- `uv run ruff format`
- `uv run ruff check`
- `uv run ty check`
- `uv run pytest` (1323 passed)
- Before the fix, new regression tests failed due to the absence of `cli.session.shutil`; after the fix, they pass.

[Residual Risks]
- Live smoke did not start in real Telegram on Windows: there is no Windows/Telegram credentials in the local environment. The behavior is covered by unit tests around the startup subprocess.


akashhg2007:
i need add google ai studio CLI

PR #540, opened 2026-05-23
fix(providers/opencode): forward reasoning_effort for DeepSeek V4 models

MysticCoss:
## Summary
  - When Claude Code users set `/effort`, the CLI sends `output_config.effort` and `thinking.budget_tokens` in API requests
  - The `opencode_go` provider was silently dropping these, so DeepSeek V4 Flash/Pro never received reasoning effort configuration
  - This fix extracts effort from `output_config.effort` (primary) or derives it from `thinking.budget_tokens` (fallback), maps Anthropic effort levels to DeepSeek's `reasoning_effort` values, and injects it
  at the top level of the upstream request body

  ## Effort mapping
  | Anthropic effort | DeepSeek reasoning_effort |
  |---|---|
  | `low` | `high` |
  | `medium` | `high` |
  | `high` | `max` |
  | `xhigh` | `max` |
  | `max` | `max` |

  ## Scope
  - Only activates for models matching `deepseek-v4-pro` and `deepseek-v4-flash`
  - Only when `thinking_enabled` is true
  - Non-DeepSeek providers are unaffected

  ## Test plan
  - [x] Model detection correctly identifies prefixed variants (e.g. `deepseek-ai/deepseek-v4-pro`)
  - [x] Effort extracted from `output_config.effort` when present
  - [x] Effort derived from `thinking.budget_tokens` when absent
  - [x] `output_config.effort` takes priority over `budget_tokens`
  - [x] No `reasoning_effort` injected when thinking is disabled or for non-DeepSeek models


Alishahryar1 (approved):

Alishahryar1:
I don't like the isdeepseek function I don't want the proxy to be model based at all, as you can see in related code it is all generic using the fact that fields can be ignored.

MysticCoss:
The reasoning effort config is helpful in heavy reasoning task and i dont want to give up on it.
I will try to make a converter to convert claude reasoning effort to other models reasoning effort. It must not model based

PR #553, opened 2026-05-25
feat(providers): per-provider BASE_URL env + PROVIDER_EXTRA_HEADERS injection

dream-ai-corp:
## Why

Some downstream users need to route FCC's upstream provider traffic through:
- **Observability proxies** (LiteLLM, Helicone, LangFuse, Portkey, …) that meter / attribute usage per tenant or feature
- **On-prem mirrors** of the cloud providers (e.g. self-hosted NVIDIA NIM)
- **Regional / compliance endpoints**

Today only the *local* providers (LM Studio, llama.cpp, Ollama) expose a base-URL env override, and there is no way to attach attribution headers to upstream requests.

This PR adds both, without changing default behavior.

### 1. `<PROVIDER>_BASE_URL` env vars (13 cloud providers)

`NVIDIA_NIM_BASE_URL`, `OPENROUTER_BASE_URL`, `GEMINI_BASE_URL`, `DEEPSEEK_BASE_URL`, `MISTRAL_BASE_URL`, `CODESTRAL_BASE_URL`, `OPENCODE_BASE_URL`, `OPENCODE_GO_BASE_URL`, `WAFER_BASE_URL`, `KIMI_BASE_URL`, `CEREBRAS_BASE_URL`, `GROQ_BASE_URL`, `FIREWORKS_BASE_URL`.

Empty default ⇒ falls back to the provider's built-in upstream URL.

**Z.ai stays intentionally pinned** — the existing `test_zai_descriptor_uses_fixed_cloud_base_url` guard is preserved, and the catalog has a comment explaining why `ZAI_BASE_URL` is not exposed.

### 2. `PROVIDER_EXTRA_HEADERS` env var

JSON-encoded dict of HTTP headers attached to every upstream provider request — both OpenAI-compatible `/chat/completions` (via `AsyncOpenAI(default_headers=...)`) and native Anthropic `/messages` (via the httpx client's default headers).

```bash
PROVIDER_EXTRA_HEADERS='{"x-tenant-id":"team-a","x-feature":"chat"}'
```

Empty / unset ⇒ no extra headers, `default_headers=None`, identical to today's behavior.

## Wiring

- `config.settings.Settings`: 13 new `*_base_url` fields + one `provider_extra_headers` field.
  Uses `Annotated[..., NoDecode]` so the validator (not pydantic-settings' eager JSON decoder) owns the parsing — empty string → `{}`, invalid JSON → clear error, non-object JSON → clear error.
- `config.provider_catalog.ProviderDescriptor`: each cloud descriptor now declares its `base_url_attr`, picked up by the existing `build_provider_config` machinery — no registry refactor required.
- `providers.base.ProviderConfig`: new `extra_headers: dict[str, str]`, defaults empty.
- `providers.openai_compat.OpenAIChatTransport`: passes `default_headers` to `AsyncOpenAI`. `None` when empty to keep current behavior bit-for-bit.
- `providers.anthropic_messages.AnthropicMessagesTransport`: passes `headers` to the httpx client. Per-request headers from subclasses still override on conflicts (httpx merge semantics).

## Tests

`tests/providers/test_extra_headers_and_base_url.py` — 41 cases:

- Settings env loading for every new `*_BASE_URL` (parametrized across 13 providers).
- `PROVIDER_EXTRA_HEADERS`: empty, valid JSON, type coercion, invalid JSON, non-object JSON.
- Z.ai pinning preserved (`hasattr(settings, "zai_base_url")` is False).
- Descriptor `base_url_attr` wiring parametrized across all 13 cloud providers.
- `build_provider_config` overrides default base URL and propagates `extra_headers`; mutating the resulting dict does not leak back into Settings.
- OpenAI-compat transport receives `default_headers=...` (or `None` when unset).
- Anthropic-Messages transport receives `headers=...` (or `None` when unset).
- `ProviderConfig.extra_headers` default is `{}`.

## CI

Following AGENTS.md:

| Check | Result |
|---|---|
| `uv run ruff format` | clean |
| `uv run ruff check` | All checks passed |
| `uv run ty check` | All checks passed |
| `uv run pytest` | **1470 passed** |

No `# type: ignore` / `# ty: ignore` introduced.

## Backwards compatibility

Zero breaking change. All defaults preserve current behavior:
- Cloud `*_BASE_URL` unset → existing default URL used.
- `PROVIDER_EXTRA_HEADERS` unset → no headers passed (`default_headers=None`).
- Z.ai pin preserved.
- Existing 1429 tests pass unchanged.

## Use case driving this PR

Routing FCC through a [LiteLLM-based metering proxy](https://github.com/BerriAI/litellm) so cost / volume per `(tenant, feature)` can be attributed back to the FCC instance — without forking FCC or sniffing traffic at the network layer.


Alishahryar1:
This is not needed for the purposes of this repo

PR #554, opened 2026-05-25
Add mid-stream transport-error retry with configurable TRANSPORT_MAX_RETRIES

marioja:
## Problem

NVIDIA NIM (and other OpenAI-compatible providers) can drop the connection mid-stream after sending a `200 OK` and starting the chunked SSE response. This manifests as `httpx.RemoteProtocolError` ("peer closed connection") or `httpx.ReadError` — provider-side failures that fcc-server previously surfaced to the client as-is, with no retry attempt.

## How to reproduce

1. Configure fcc-server to use NVIDIA NIM as the provider (`MODEL=nvidia_nim/...`)
2. Send a long streaming request (e.g. a complex code generation prompt)
3. Occasionally the NIM server drops the connection before the chunked response completes
4. Client receives an error like: `RemoteProtocolError: peer closed connection`

This is intermittent and provider-side — not caused by fcc-server or the client.

## What this fixes

**Commit 1: `Add mid-stream transport-error retry for RemoteProtocolError`**
- Wraps the entire `_stream_response_impl` streaming loop in a retry mechanism (up to 2 retries by default)
- On `RemoteProtocolError` or `ReadError` with retries remaining: emits a clean `message_stop` for the failed attempt, sleeps 1s, then starts a fresh `SSEBuilder` on the next attempt
- On exhausted retries: prefixes the error message with `fcc-server X retry ...` so the caller knows retries were attempted
- Non-retryable errors (e.g. `ReadTimeout`, `RuntimeError`) surface immediately without retry or prefix
- Because this is an async generator, we cannot "un-yield" SSE events already sent — each retry iteration creates a fresh builder so the client sees at most one complete message

**Commit 2: `Make TRANSPORT_MAX_RETRIES configurable via env var`**
- Replaces the hardcoded `_TRANSPORT_MAX_RETRIES` class attribute with a configurable `TRANSPORT_MAX_RETRIES` env var
- Wires through the full config chain: `constants.py` → `settings.py` → `providers/base.py` → `registry.py` → `openai_compat.py`
- Adds admin UI field (`ConfigFieldSpec` in `admin_config.py`) under Runtime (advanced)
- Documents in `.env.example`
- Set `TRANSPORT_MAX_RETRIES=0` to disable retry, or increase for flaky providers

## Files changed (9)

| File | Change |
|---|---|
| `config/constants.py` | Add `TRANSPORT_MAX_RETRIES_DEFAULT = 2` |
| `config/settings.py` | Add `transport_max_retries` field with `TRANSPORT_MAX_RETRIES` env alias |
| `providers/base.py` | Add `transport_max_retries` to `ProviderConfig` |
| `providers/registry.py` | Pass `settings.transport_max_retries` into `ProviderConfig()` |
| `providers/openai_compat.py` | Replace `_TRANSPORT_MAX_RETRIES` class attr with `self._config.transport_max_retries`; add retry loop in `_stream_response_impl` |
| `api/admin_config.py` | Register `TRANSPORT_MAX_RETRIES` as advanced runtime field |
| `.env.example` | Document `TRANSPORT_MAX_RETRIES=2` |
| `tests/providers/test_openai_compat_transport_retry.py` | 12 tests: retry once/twice/exhausted, ReadError retryable, non-retryable, mid-stream retry, zero retries, custom count |
| `docs/testing-transport-retry.md` | Instructions for running from source to test |

## Test plan

- [x] 12 unit tests covering: retry-once-and-succeed, retry-twice-and-succeed, exhausted-retries-prefix, ReadError retryable, non-retryable no prefix, mid-stream retry, exhausted after partial, ReadTimeout non-retryable, zero-retries-disables-retry, custom-retry-count, custom-retry-exhausted, default-is-two
- [x] Config chain smoke tested: `TRANSPORT_MAX_RETRIES` env var → `Settings` → `ProviderConfig` → `openai_compat.py`
- [x] AST parse verified on all 7 changed source files
- [ ] Manual test: run from source (`.venv/Scripts/python.exe -m cli.entrypoints`) and verify retry log lines appear on NIM connection drops
- [ ] Admin UI: confirm TRANSPORT_MAX_RETRIES appears under Runtime (advanced) at `/admin`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -133,6 +133,9 @@ HTTP_READ_TIMEOUT=300
 HTTP_WRITE_TIMEOUT=60
 HTTP_CONNECT_TIMEOUT=60
 
+# Mid-stream transport error retries (0 disables retry)
+TRANSPORT_MAX_RETRIES=2
+
 
 # Optional server API key (Anthropic-style)
 ANTHROPIC_AUTH_TOKEN="freecc"
--- a/api/admin_config.py
+++ b/api/admin_config.py
@@ -527,6 +527,16 @@ FIELDS: tuple[ConfigFieldSpec, ...] = (
         default="60",
     ),
     ConfigFieldSpec(
+        "TRANSPORT_MAX_RETRIES",
+        "Transport Error Retries",
+        "runtime",
+        "number",
+        settings_attr="transport_max_retries",
+        default="2",
+        advanced=True,
+        description="Max mid-stream transport-error retries before surfacing error (0 disables retry).",
+    ),
+    ConfigFieldSpec(
         "HOST",
         "Server Host",
         "runtime",
--- a/config/constants.py
+++ b/config/constants.py
@@ -3,6 +3,9 @@
 # HTTP client connect timeout (seconds). Keep aligned with README.md and .env.example.
 HTTP_CONNECT_TIMEOUT_DEFAULT = 10.0
 
+# Mid-stream transport error retries (provider drops connection during chunked response).
+TRANSPORT_MAX_RETRIES_DEFAULT = 2
+
 # Anthropic Messages API default when the client omits max_tokens.
 ANTHROPIC_DEFAULT_MAX_OUTPUT_TOKENS = 81920
 
--- a/config/settings.py
+++ b/config/settings.py
@@ -11,7 +11,7 @@ from dotenv import dotenv_values
 from pydantic import Field, field_validator, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-from .constants import HTTP_CONNECT_TIMEOUT_DEFAULT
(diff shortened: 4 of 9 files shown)


marioja:
I use this all the time, works great.  Haven't seen a connection drop from a recoverable error since running it.  It should show a message in fcc-server console.  I don't check it carefully but I have not seen one.

Alishahryar1:
I'm closing this PR and doing it myself since it's not correct

marioja:
> I'm closing this PR and doing it myself since it's not correct

That's cool. what did you not like about it? just for improving it for next time? Missing code for other providers?  @Alishahryar1 Let me know once you have implemented it.

Alishahryar1:
The implementation is done Check it out in one of the recent commits, it has to do with mid stream recovery, if we retry the whole thing, we will see duplicate messages

PR #572, opened 2026-05-27
fcc-claude: manage the proxy lifecycle and stop it on exit

ovuruska:
## Summary

`fcc-claude` previously required a separately running `fcc-server` and, when one was running, **left it running after the session**. In non-interactive (`-p`) runs this is clearly a bug: the one-shot command finishes but the proxy lingers with nothing to serve (you have to hunt down and kill it manually).

This change makes `fcc-claude` own the proxy lifecycle.

## Behavior

- **Auto-start:** if nothing answers `/health`, `fcc-claude` starts a proxy it owns — browser auto-open suppressed, logs teed to `~/.fcc/managed-server.log` — waits for it to become healthy, then launches Claude Code.
- **Auto-stop:** on exit (normal, `-p` one-shot, error, or Ctrl-C) the proxy we started is gracefully stopped (SIGTERM → wait → force-kill the tree).
- **Reuse, don't hijack:** a proxy that is **already running** (e.g. a separate `fcc-server`) is reused and **left untouched** on exit — so concurrent sessions and long-lived servers are safe. We only ever stop a proxy this CLI started.
- **Opt-out:** `FCC_AUTO_SERVER=0` restores the previous behavior (print a hint and exit instead of auto-starting).

`launch_claude` is split into `_ensure_proxy_running` + `_run_claude_client`; the client-spawn path keeps its exact prior behavior and exit codes.

## Tests

New unit coverage in `tests/cli/test_entrypoints.py`: auto-start + teardown, reuse-without-stopping, `FCC_AUTO_SERVER` parsing, managed `serve` command construction (+ browser suppression), readiness early-exit detection, and force-kill on shutdown timeout. The pre-existing "unreachable proxy" test now exercises the `FCC_AUTO_SERVER=0` opt-out path.

## Validation

- `ruff format`, `ruff check`, `ty check` — clean.
- `uv run pytest` — **1435 passed**.
- **Verified live:** `fcc-claude -p` with no server running auto-starts the proxy, returns an answer, and on exit leaves port `8082` closed with no lingering process (`managed-server.log` shows a clean `Finished server process`).

### DIFF
--- a/cli/entrypoints.py
+++ b/cli/entrypoints.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import os
 import shutil
 import subprocess
@@ -30,6 +31,8 @@ from config.settings import Settings, get_settings
 PROXY_PREFLIGHT_PATH = "/health"
 PROXY_PREFLIGHT_TIMEOUT_SECONDS = 1.5
 SERVER_GRACEFUL_SHUTDOWN_SECONDS = 5
+SERVER_READY_TIMEOUT_SECONDS = 30.0
+SERVER_READY_POLL_INTERVAL_SECONDS = 0.15
 
 
 def _load_env_template() -> str:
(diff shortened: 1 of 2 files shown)


Alishahryar1:
I don't know about this one, server is supposed to be separate from Claude. Closing for that reason.

PR #575, opened 2026-05-27
fix(gemini): use reasoning_effort for thinking control on Gemini 3.x

scs640:
## Summary
- Replaces the double-nested `extra_body.google.thinking_config` approach with the OpenAI-standard `reasoning_effort` parameter, per Google's documented Gemini 3.x mappings
- The two parameters are mutually exclusive, and omitting `thinking_config` keeps `include_thoughts` off by default -- preventing the model's internal reasoning from leaking as visible text in chat
- Defaults `reasoning_effort` to `"low"` to reduce over-engineering for simple prompts

## Changes
- Map Anthropic `budget_tokens` to `reasoning_effort` (minimal/low/medium/high)
- Strip `temperature`/`top_p` when thinking is enabled (interferes with reasoning)
- Inject `thought_signatures` on tool calls (required by Gemini 3 models)
- Set `parallel_tool_calls` to false
- Add defensive `or 0` guard for tool call index in `openai_compat.py`

## Test plan
- [x] `uv run pytest tests/providers/test_gemini.py -v` -- 22/22 pass
- [x] Verified `reasoning_effort` mappings match Google's documented Gemini 3.x behavior
- [x] Confirmed clean diff: no `.DS_Store`, no unrelated debug logging


scs640:
New commit submitted to resolve Greptile comments, but it has not re-triggered Greptile.

PR #577, opened 2026-05-27
Fix Gemini model validation warning and tool call index crash

poluriha:
## Summary

Two bugs found while debugging `fcc-claude start` returning `Invalid request sent to provider`.

### Bug 1 - Startup WARNING: `missing model` (`providers/gemini/client.py`)

Gemini's OpenAI-compatible `/models` endpoint returns model IDs with a `models/` prefix (e.g. `models/gemini-2.5-flash`), but the configured model in `.env` is `gemini/gemini-2.5-flash` (resolved to `gemini-2.5-flash` after stripping the provider prefix). The validator compared `gemini-2.5-flash` against `models/gemini-2.5-flash` and found no match, producing the startup warning.

**Fix:** Override `list_model_infos()` in `GeminiProvider` to strip the `models/` prefix from all model IDs returned by the API.

### Bug 2 - `TypeError: '<' not supported between instances of 'NoneType' and 'int'` (`providers/openai_compat.py`)

When Gemini returns a streaming tool call chunk with `"index": null`, `tc.get("index", 0)` returns `None` - Python's `dict.get(key, default)` only uses the default when the key is **absent**, not when the value is `None`. The subsequent `if tc_index < 0` comparison then raises a `TypeError`.

**Fix:** Replace `tc.get("index", 0)` with an explicit `None`-check: `tc.get("index") if tc.get("index") is not None else 0`.

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/providers/gemini/client.py
+++ b/providers/gemini/client.py
@@ -6,6 +6,7 @@ from typing import Any
 
 from providers.base import ProviderConfig
 from providers.defaults import GEMINI_DEFAULT_BASE
+from providers.model_listing import ProviderModelInfo
 from providers.openai_compat import OpenAIChatTransport
 
 from .request import build_request_body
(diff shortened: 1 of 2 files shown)


poluriha:
Hey @Alishahryar1, could you approve the CI workflow run when you get a chance? It's blocked on first-time contributor approval. Thanks!

poluriha:
Closing this since I've moved to maintaining my own fork at https://github.com/poluriha/free-claude-code with these fixes applied. Feel free to pick them up if useful!

PR #592, opened 2026-05-28
fix: default model nvidia_nim/z-ai/glm4.7 does not exist on NVIDIA NIM

jalexiscv:
The hardcoded default model `nvidia_nim/z-ai/glm4.7` is not a valid NVIDIA NIM model. Users who start without configuring `.env` get `"Invalid request sent to provider"` errors.

The `.env.example` recommends `nvidia_nim/nvidia/nemotron-3-super-120b-a12b` which is a valid NIM model. Updated the default to match.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ debug-*.log
 llama_cache
 .smoke-results
 .vscode
-server.*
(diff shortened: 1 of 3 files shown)


Alishahryar1:
Fixed already

PR #639, opened 2026-05-30
fix: accept system-role messages and fold into top-level system field

smellyborelli:
## Summary

Claude Code sends mid-conversation system prompt updates as `role: "system"` messages in the messages array (context-management edits). The current Pydantic `Message` model rejects these at validation time, returning HTTP 422 before the request reaches any provider.

```
Request validation failed: path=/v1/messages error_locs=[['body', 'messages', 1, 'role']]
message_summary=[{'role': 'user', ...}, {'role': 'system', 'content_kind': 'str', 'content_length': 22119}]
```

## Changes

- **`api/models/anthropic.py`** — allows `"system"` in the `Message.role` Literal so validation passes
- **`providers/deepseek/request.py`** — adds `_fold_system_messages_to_top_level()` which extracts system-role messages from the messages array and appends their content to the top-level `system` field (proper Anthropic format). Called before forwarding to DeepSeek, whose API rejects system-role messages in the conversation array
- **`core/anthropic/conversion.py`** — re-routes mid-conversation system messages as user messages in the OpenAI converter path (OpenAI also rejects system-role messages after the first position)

## Test plan

- [x] Verified proxy starts and serves `/v1/models` after changes
- [x] Existing conversation flow through the proxy continues to work
- [x] No change to behavior for requests without system-role messages (early-return in `_fold_system_messages_to_top_level`)

### DIFF
--- a/api/models/anthropic.py
+++ b/api/models/anthropic.py
@@ -142,7 +142,7 @@ def _normalize_system_role_messages(data: Any) -> Any:
 # Message Types
 # =============================================================================
 class Message(BaseModel):
-    role: Literal["user", "assistant"]
+    role: Literal["user", "assistant", "system"]
     content: (
         str
         | list[
--- a/core/anthropic/conversion.py
+++ b/core/anthropic/conversion.py
@@ -214,6 +214,19 @@ class AnthropicToOpenAIConverter:
                             reasoning_replay=reasoning_replay,
                         )
                     )
+            elif role == "system":
+                # Mid-conversation system messages (context-management edits) cannot be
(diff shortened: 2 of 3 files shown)


mohamedibourki (commented):
Review note: lgtm on the approach — folding system-role messages into the top-level system field matches Claude Code’s context-management behavior. Verify the downstream providers that reject nested system messages (including NIM paths) accept the rewritten body before merging, otherwise some callers may still see 422s. Otherwise no blockers from my side.

Alishahryar1:
Fixed myself

Issue #654, opened 2026-05-31
'<' not supported between instances of 'NoneType' and 'int' (request_id=req_(numbers))


ssaannzzhhiikk:
This problem always appears when i try to do smth how to handle it?

Alishahryar1:
fixed

PR #657, opened 2026-05-31
fix: guard rate_limit=None in GlobalRateLimiter initialization (#654)

mohamedibourki:
Fixes the TypeError when rate_limit=None reaches the constructor.

The bug: GlobalRateLimiter.__init__ could receive rate_limit=None (for example via config.map() or provider.get_scoped_instance(...)) and crash on the 'rate_limit <= 0' comparison with a misleading TypeError instead of a clean ValueError.

This change makes rate_limit=None resolve to the built-in default (40) before any comparison, so the validation path always sees an int.

Addresses reviewer feedback: updated the regression test to call GlobalRateLimiter(...) directly, bypassing get_instance, so the new None-guard in __init__ is actually exercised by pytest.

Changes
- providers/rate_limit.py: accept rate_limit: int | None = None, resolve None to 40 before comparing
- tests/providers/test_provider_rate_limit.py: regression test exercises the constructor directly

Verification
- uv run pytest tests/providers/test_provider_rate_limit.py: 35 passed
- ruff-format, ruff-check, ty: clean

### DIFF
--- a/providers/rate_limit.py
+++ b/providers/rate_limit.py
@@ -62,7 +62,7 @@ class GlobalRateLimiter:
 
     def __init__(
         self,
-        rate_limit: int = 40,
+        rate_limit: int | None = None,
         rate_window: float = 60.0,
         max_concurrency: int = 5,
     ):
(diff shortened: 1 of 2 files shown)


mohamedibourki (commented):
Addressed the earlier reviewer note: the regression test now calls GlobalRateLimiter(rate_limit=None, ...) directly rather than going through get_instance(...), so the new None->40 branch is actually exercised. Re-check / re-run when convenient, otherwise no blockers from my side. Thanks.

mohamedibourki:
Re-check requested: reviewer note is addressed (regression test now hits __init__ directly). All local checks (ruff format, ruff check, ty check, pytest) pass. Ready for re-review whenever convenient.

PR #700, opened 2026-06-03
feat(nvidia_nim): round-robin key rotation, XML tool extraction, stream resilience

FranciscoDanielMunguiaZertuche:
## Summary

This PR adds several production-hardening improvements to the NVIDIA NIM provider, along with core tool parsing and SSE resilience enhancements that benefit all OpenAI-compatible providers.

### NVIDIA NIM Provider

- **Round-robin API key rotation**: Accepts comma-separated API keys in `NVIDIA_NIM_API_KEY`. On 429 rate limit or read timeout, automatically cycles to the next key before retrying.
- **Progressive 400-error retry**: Strips NIM-specific fields in order of decreasing specificity (`reasoning_budget` -> `chat_template` -> `reasoning_content` -> all thinking fields), so requests that fail due to unsupported features get progressively downgraded rather than failing outright.
- **Overridden `_create_stream`**: Integrates key rotation directly into the stream creation path so retryable errors trigger key cycling transparently.

### Core Tool Parsing

- **XML tool call extraction** (`HeuristicToolParser.extract_xml_tool_calls`): NIM models with `thinking=True` emit tool calls as XML inside `reasoning_content` instead of structured `tool_calls`. This stateless method detects `<ToolName><param>value</param></ToolName>` patterns and converts them to proper `tool_use` blocks.
- **Smart parameter cleaning** (`_clean_xml_parameter`): Preserves multi-line code content (e.g., `oldText`/`newText` in Edit calls) while stripping XML formatting artifacts from single-line fields.
- **Tool name validation**: Suppresses phantom tool calls where the model hallucinates a tool name not in the request -- emits as text instead of a broken `tool_use` block that crashes Claude Code.

### SSE Stream Resilience

- **`_rescue_partial_json`**: When a transport error interrupts a stream mid-tool-call, completes the partial JSON so the client can parse it. Handles unterminated strings, open braces/brackets, and falls back to `{"_interrupted": true}`.
- **`close_incomplete_tool_blocks`**: Pre-close validation that repairs any tool blocks with invalid accumulated JSON.
- **`_strip_markdown_fences`**: Handles models that wrap tool JSON in code fences.
- **Empty tool_id/name guard**: Prevents `content_block_start` with empty identifiers.

### Configuration

- `MODEL_VISION`: Route image-bearing requests to a vision-capable model.
- `NIM_STRIP_TOOLS`: Comma-separated tool names to strip from NIM requests.
- `DEEPSEEK_BASE_URL`: Custom DeepSeek endpoint support.
- Increased default `HTTP_READ_TIMEOUT` from 120s to 300s for NIM compatibility.
- `MidStreamDisconnectError` exception for structured mid-stream error handling.

### Testing

All 1484 tests pass, including new tests for:
- XML tool call extraction and parameter cleaning
- SSE builder rescue methods
- NIM provider round-robin and retry logic

### Why These Changes Matter

The NIM provider is the default in this project, and these changes make it significantly more robust in production:

1. Multi-key rotation prevents cascading failures when one key hits rate limits
2. Progressive retry means requests that would fail outright on NIM-specific quirks get silently downgraded
3. XML tool parsing handles a real NIM platform behavior where structured tool_calls aren't emitted
4. Stream rescue prevents "tool call could not be parsed" errors that break Claude Code sessions

These improvements are backward-compatible -- existing single-key configurations work unchanged, and the new features only activate when the relevant environment variables are set.

### DIFF
--- a/api/services.py
+++ b/api/services.py
@@ -149,7 +149,15 @@ class ClaudeProxyService:
                 return optimized
             logger.debug("No optimization matched, routing to provider")
 
-            provider = self._provider_getter(routed.resolved.provider_id)
+            # Prefer passing `claude_model` kwarg but fall back to older/test
+            # callables that accept only the provider_id positional argument.
+            try:
+                provider = self._provider_getter(
+                    routed.resolved.provider_id,
+                    claude_model=routed.resolved.original_model,
+                )
+            except TypeError:
+                provider = self._provider_getter(routed.resolved.provider_id)
             provider.preflight_stream(
                 routed.request,
                 thinking_enabled=routed.resolved.thinking_enabled,
--- a/config/nim.py
+++ b/config/nim.py
@@ -35,6 +35,11 @@ class NimSettings(BaseModel):
     min_tokens: int = Field(0, ge=0, description="Minimum tokens in the response.")
     chat_template: str | None = None
     request_id: str | None = None
+    reasoning_effort: str | None = None
+    strip_tools: list[str] = Field(
+        default_factory=list,
+        description="Tool names to strip from requests (e.g. image tools that confuse the model).",
+    )
 
     model_config = ConfigDict(extra="forbid")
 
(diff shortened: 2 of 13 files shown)


Alishahryar1:
Limits are enforced per account not per key

bozoweed:
nice violating ToS ..... maye be should i report that to NIM support and hopping you will be banned from NIM <3

PR #1203, opened 2026-07-19
Add OpenAI Chat Completions endpoint (/v1/chat/completions)

wromansky:
## Problem

| Before | After |
| --- | --- |
| The proxy served the Anthropic Messages (`/v1/messages`) and OpenAI Responses (`/v1/responses`) dialects, but **not** OpenAI Chat Completions. | The proxy also serves a native **`POST /v1/chat/completions`** (with a `HEAD`/`OPTIONS` probe), so any OpenAI chat-completions client can use it directly. |
| Clients that only speak chat-completions — e.g. **jcode's `openai-compatible` provider profile**, LiteLLM-style routers — got `404` and could not connect. | Those clients connect unchanged: point them at `<base>/v1` with the proxy bearer token. |

## Approach

The Chat Completions handler mirrors the existing Responses flow. It converts the incoming chat request into the shared internal Anthropic `MessagesRequest`, runs it through the **same** provider pipeline (`ProviderExecutor`), and re-serializes the result in the Chat Completions dialect:

- **non-streaming** → aggregate the internal SSE via the existing `aggregate_anthropic_sse_to_message`, then emit a single `chat.completion`;
- **streaming** → translate the Anthropic SSE into `chat.completion.chunk` frames (role delta, content deltas, `tool_calls` deltas, `finish_reason`, `[DONE]`).

Coverage: `system`/`user`/`assistant`/`tool` roles, `tool_calls` ↔ Anthropic `tool_use`, `tool_choice` (`auto`/`none`/`required`/named), multimodal `image_url` parts, `stop` sequences, and `stream_options.include_usage`. Anthropic `stop_reason` maps to `finish_reason` (`end_turn`/`stop_sequence`→`stop`, `max_tokens`→`length`, `tool_use`→`tool_calls`).

### Changes
- New `core/openai_chat/` adapter package (request model, input conversion, tool/tool_choice conversion, streaming assembler, non-stream completion builder, finish-reason mapping, SSE formatting).
- New `api/handlers/chat_completions.py` + route + `HEAD`/`OPTIONS` probe, wired through `ApiServices` exactly like the other dialects.
- `wire_api="chat"` trace labels in `application/execution.py`; the shared OpenAI error envelope owned by `core/openai_responses` is reused (as `api/request_errors` already does).
- **`ARCHITECTURE.md` updated** per the "Maintenance Rules" (this adds a public route + wire protocol): the routes list, a new `core/openai_chat/` protocol-conversion section mirroring the Responses one, and the "Add Protocol Behavior" extension checklist.
- Tests for request conversion, streaming chunks, non-stream aggregation, and the HTTP endpoint (stream, non-stream, tool calls, auth-required, preflight rejection).
- MINOR version bump `4.11.0` → `4.12.0` with matching `uv.lock`.

## Verification
- `./scripts/ci.sh`: ruff-format, ruff-check, ty, and pytest all pass (the new suite adds 25 tests). *(The only failures in my environment are the pre-existing `tests/scripts/` installer tests, which also fail on a clean `main` checkout here.)*
- **Live, end-to-end against real providers** (running the branch server, driving it with `curl` and with jcode):
  - `GET /openapi.json` lists `POST /v1/chat/completions`; unauthenticated calls return `401`.
  - Non-streaming completion through NVIDIA NIM returns a well-formed `chat.completion` (correct `finish_reason`, `usage`).
  - Streaming completion returns ordered `chat.completion.chunk` frames ending in `[DONE]`, with a correct trailing usage chunk under `stream_options.include_usage`.
  - **jcode's `openai-compatible` provider profile** (`jcode provider add ... --base-url http://HOST/v1`) runs a full multi-turn agentic session (including tool calls) through the endpoint — the exact client that could not connect before.

## Residual Risks
- Thinking/reasoning content blocks are intentionally dropped from the Chat Completions output (that dialect has no standard field for them); text and tool calls are unaffected.
- The shared OpenAI error envelope now has two dialect consumers; a future cleanup could hoist it into a neutral `core/openai_shared` module (kept out of scope here to keep the diff focused).

<h3>Greptile Summary</h3>

This PR adds native OpenAI Chat Completions support to the proxy. The main changes are:

- Adds authenticated `POST`, `HEAD`, and `OPTIONS` handling for `/v1/chat/completions`.
- Converts Chat Completions requests into the shared Anthropic Messages pipeline.
- Translates streaming and non-streaming provider output into OpenAI-compatible responses.
- Supports tools, tool choices, multimodal input, stop sequences, and usage chunks.
- Adds endpoint and adapter tests, architecture documentation, and the 4.12.0 version update.

This looks safe to merge.

The updated stream framing handles CRLF events. Incomplete content-bearing streams no longer report normal completion without a completion signal. Tool and tool-choice validation now rejects the malformed cases covered by the original fixes. No additional publishable blocking issue was found in the updated code.

**What T-Rex did**
- Compared pre-change behavior and confirmed that POST /v1/chat/completions returned 404 Not Found before the revision.
- Verified post-change exposure: OpenAPI now supports POST, HEAD, and OPTIONS; probes returned 204, and non-stream, streaming, and tool-call requests returned 200, with invalid preflight yielding a 400 envelope.
- Collected streaming evidence showing ordered frames including role, content, stop, usage, and \[DONE\].
- Diagnostic runner and pytest completed successfully with exit code 0.

<a href="https://app.greptile.com/trex/runs/15034300/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| src/free_claude_code/core/openai_chat/streaming.py | Translates Anthropic SSE events into Chat Completions chunks and handles framing, termination, errors, tool calls, and usage. |
| src/free_claude_code/core/openai_chat/input.py | Converts Chat Completions messages, tools, images, and generation options into Anthropic request payloads. |
| src/free_claude_code/api/handlers/chat_completions.py | Runs chat requests through provider execution and serializes streaming, aggregated, and error responses. |
| src/free_claude_code/api/routes.py | Registers the authenticated Chat Completions endpoint and preserves request-runtime lease ownership. |

<sub>Reviews (2): Last reviewed commit: ["Add OpenAI Chat Completions endpoint (/v..."](https://github.com/alishahryar1/free-claude-code/commit/8d177e0146345483a3a8ed67456deefe875ce3c6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45489776)</sub>

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -351,6 +351,7 @@ and non-local origins.
 
 - `POST /v1/messages`: Anthropic Messages-compatible streaming requests.
 - `POST /v1/responses`: OpenAI Responses-compatible requests.
+- `POST /v1/chat/completions`: OpenAI Chat Completions-compatible requests.
 - `POST /v1/messages/count_tokens`: Anthropic token counting.
 - `GET /v1/models`: gateway and Claude-compatible model listing.
 - `GET /health`: health check.
@@ -860,6 +861,31 @@ does not synthesize reasoning summaries.
 Provider code should delegate protocol details to these modules. Avoid copying
 conversion code into individual providers, and avoid provider-to-provider imports
 for shared Anthropic behavior.
+
+[src/free_claude_code/core/openai_chat/](src/free_claude_code/core/openai_chat/) owns OpenAI Chat
+Completions support and mirrors the Responses dialect:
+
+- the permissive `OpenAIChatCompletionsRequest` ingress model used directly by
+  the FastAPI route and the protocol adapter;
+- the `OpenAIChatAdapter` facade used by the API layer;
+- both non-streaming and streaming `/v1/chat/completions` support;
+- Chat request conversion into Anthropic Messages payloads, covering
+  `system`/`user`/`assistant`/`tool` roles, `tool_calls` to `tool_use`,
+  `tool_choice`, multimodal `image_url` parts, and stop sequences;
+- non-streaming responses folded from the internal Anthropic SSE through the
+  shared `aggregate_anthropic_sse_to_message` and serialized as a single
+  `chat.completion`;
+- streaming responses assembled as `chat.completion.chunk` frames by
+  `streaming.py`, reusing the neutral Anthropic SSE parser under
+  `core/anthropic/`.
+
+Anthropic `stop_reason` maps to Chat `finish_reason` (`tool_use` to
+`tool_calls`, `max_tokens` to `length`, otherwise `stop`), and
+`stream_options.include_usage` appends a trailing usage-only chunk. The package
+reuses the OpenAI error envelope owned by `core/openai_responses`. Reasoning and
+thinking blocks are dropped from Chat Completions output because that dialect has
+no standard field for them; text and tool calls are unaffected. API code depends
+on the adapter, not on the internal module owners.
 
 ## Local Optimizations And Server Tools
 
@@ -1391,7 +1417,8 @@ when maintainers want branch-level assurance.
 
 1. Put shared Anthropic behavior under [src/free_claude_code/core/anthropic/](src/free_claude_code/core/anthropic/).
 2. Put OpenAI Responses behavior under
-   [src/free_claude_code/core/openai_responses/](src/free_claude_code/core/openai_responses/).
+   [src/free_claude_code/core/openai_responses/](src/free_claude_code/core/openai_responses/), and OpenAI
+   Chat Completions behavior under [src/free_claude_code/core/openai_chat/](src/free_claude_code/core/openai_chat/).
 3. Keep provider-specific request quirks inside the provider profile or specialized
    provider subclass.
 4. Add stream contract tests under [tests/contracts/](tests/contracts/) or
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "free-claude-code"
-version = "4.11.0"
+version = "4.12.0"
 description = "Local proxy connecting coding agents to OpenAI-compatible AI providers"
 readme = "README.md"
 requires-python = ">=3.14.0"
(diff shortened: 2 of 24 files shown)


wromansky:
Thanks for the review — addressed all four findings (pushed as an amend to the same commit), with tests for each:

- **Malformed tools disappear silently** (`core/openai_chat/tools.py`): `convert_tools` now *raises* on malformed entries (non-dict, unsupported `type`, missing `function`/`name`) instead of skipping them, so the provider never receives fewer tools than the caller declared. This matches the strict validation in `core/openai_responses/tools.py`.
- **Invalid tool choice becomes auto** (`core/openai_chat/tools.py`): `convert_tool_choice` now raises on unknown strings and malformed named choices instead of returning `None`/falling back to auto. It is also validated unconditionally in `input.py` (even with no tools), matching the Responses dialect.
- **CRLF events stay buffered** (`core/openai_chat/streaming.py`): the streaming framing normalizes CRLF (`\r\n` → `\n`) before splitting on `\n\n`, so events parse under either LF or CRLF framing.
- **Truncated streams report success** (`core/openai_chat/streaming.py`): added `finish_incomplete_stream`. If the provider stream ends mid-content with no `message_stop` **and** no `message_delta.stop_reason`, it emits an error frame instead of a fabricated successful completion. A `stop_reason` seen without a trailing `message_stop` still finishes normally, and an empty stream is unchanged (matches Responses).

Local CI (`./scripts/ci.sh`: ruff-format, ruff-check, ty, pytest) is green; the new suite adds coverage for strict validation, CRLF framing, and the truncation guard. (The only failing tests in my environment are the pre-existing `tests/scripts/` installer tests, which also fail on a clean `main` checkout here.)

Alishahryar1:
I don't plan on supporting this

wromansky:
👍.

PR #1292, opened 2026-07-30
feat: add fcc-claude-desktop launcher with picker aliasing (PR1)

Gakpey:
## Summary
- Add `fcc-claude-desktop` entry point for Claude Desktop Electron app
- Implement picker aliasing (`claude-sonnet-nim-####`) to work around upstream model picker limitations
- Auto-configure Claude Desktop via JSON merge in install scripts (no jq/sed — Python helper)
- Reverse merge on uninstall

## Components
| Layer | Files |
|-------|-------|
| Core alias | `src/free_claude_code/core/gateway_model_ids.py` |
| API catalog/routing | `api/model_catalog.py`, `application/routing.py` |
| Runtime seeding | `runtime/application.py` |
| CLI launcher | `cli/launchers/claude_desktop.py` (new) |
| Install/Uninstall | `scripts/install.{sh,ps1}`, `uninstall.{sh,ps1}` |
| Tests | 38 new across core/api/routing/runtime/cli |

## Validation
- 38 PR1-specific tests pass
- 2,582 total tests pass
- `ruff format` / `ruff check` / suppression ban clean
- Version bump: 4.13.1 → 4.14.0 (MINOR) + `uv.lock`

## Follow-up (PR2)
Phase 1 from integration analysis: `/v1/telemetry` 200 JSON, capability flags, billing header.

<h3>Greptile Summary</h3>

Adds Claude Desktop integration, picker-safe aliases, model catalog exposure, routing support, runtime cache seeding, and installer configuration. Runtime testing reproduced a failure after asynchronous model discovery: a newly discovered model is cached but does not receive a picker alias, so the catalog publishes canonical gateway IDs that Claude Desktop can filter out.

Merge safety: do not merge until background discovery updates picker aliases for newly added models.

<h3>Confidence Score: 2/5</h3>

**What T-Rex did**
- Executed the focused stale-alias regression script to reproduce the posted P1 finding and collect baseline and discovery outputs.
- Compared warm-cache baseline, background discovery results, and picker/catalog test outputs to validate the finding evidence.
- Validated contract behavior by reviewing the regression script results and post-discovery state, noting the model is cached with no alias and is published with canonical gateway IDs.
- Documented and linked the produced artifacts to support reviewer inspection of the regression run and its outputs.

<a href="https://app.greptile.com/trex/runs/16646739/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<h3>Comments Outside Diff (6)</h3>

1. `src/free_claude_code/runtime/application.py`, line 263 ([link](https://github.com/alishahryar1/free-claude-code/blob/adf5d66459ac7b9f018b75e8f091af72eb6545d1/src/free_claude_code/runtime/application.py#L263)) 

   <a href="#">P1</a> **Model refresh leaves picker aliases stale**

   - **Bug**
     - `start()` seeds the process-global picker alias maps from the warmed cache at line 157, but `refresh_models()` only refreshes the provider cache and returns its result. It never rebuilds those maps. `/v1/models` uses an alias only when that map contains the provider/model ref and otherwise advertises the canonical gateway wrapper ID. Consequently, a model added by a post-startup refresh is listed without the picker alias Claude Desktop accepts, so it remains unavailable in that picker until the process restarts.
   - **Cause**
     - `ApplicationRuntime.refresh_models()` delegates to `refresh_model_list_cache()` without calling `_seed_picker_aliases_from_cache()` after the cache changes.
   - **Fix**
     - After a successful model-cache refresh, rebuild picker aliases from `cached_prefixed_model_infos()` before returning the refresh result. Ensure any other supported post-startup catalog mutation path applies the same synchronization.

2. General comment 

   <a href="#">P1</a> **Auto Claude Desktop configuration calls python3 outside uv's console-entry-point bin**

   - **Bug**
     - After successfully verifying `fcc-desktop`, `fcc-server`, `fcc-claude`, `fcc-codex`, `fcc-pi`, and `fcc-claude-desktop`, the installer invokes `$tool_bin/python3 -m ...`. In a uv-style bin containing those verified entry points but no interpreter, that command returns 127. The error is treated as best-effort, so installation completes while Claude Desktop configuration is absent.
   - **Cause**
     - `uv tool dir --bin` identifies the directory for exposed tool executables; the installer assumes it also contains the tool environment's `python3`, despite only validating console scripts there.
   - **Fix**
     - Invoke the installed and already-verified console script directly: `"$tool_bin/fcc-claude-desktop" --configure`; make the dry-run output use the same command.

3. General comment 

   <a href="#">P1</a> **Claude Desktop configuration is overwritten when parsing fails or the root is not an object**

   - **Bug**
     - `load_existing_config()` returns `{}` for malformed JSON and for valid JSON values whose top level is not an object. `configure_claude_desktop_config()` then treats that empty mapping as a new config and atomically replaces the original file with the FCC configuration. The targeted run reproduced this for both `{"theme": "dark", "fontSize": 14,` and `[{"theme": "dark", "fontSize": 14}]`.
   - **Cause**
     - Lines 76-83 discard both decode failures and non-dictionary parsed values instead of preventing the write or preserving/recovering the original content.
   - **Fix**
     - Do not write when an existing configuration is malformed or has a non-object root. Return a failure/status and retain the original file (optionally create a backup and require explicit user action before replacement).

4. General comment 

   <a href="#">P1</a> **Post-startup model refresh leaves picker aliases stale**

   - **Bug**
     - `ApplicationRuntime.start()` seeds aliases from the warmed cache once, but `refresh_models()` delegates to `ProviderRuntimeManager.refresh_model_list_cache()` without reseeding. A model discovered after startup is returned by `/v1/models` with canonical/wrapper IDs instead of picker aliases.
   - **Cause**
     - `refresh_models()` has no call to `_seed_picker_aliases_from_cache()` after a successful cache refresh; aliases are process-global snapshots established only during startup.
   - **Fix**
     - After a successful model-cache refresh that changes or successfully refreshes providers, call `_seed_picker_aliases_from_cache()` before subsequent model-catalog responses. Add a regression test that starts with one model, refreshes to add another, and asserts both models receive picker aliases.

5. General comment 

   <a href="#">P1</a> **Refreshing the model cache breaks aliases already retained by Claude Desktop**

   - **Bug**
     - After a client retains a positional picker alias, `ApplicationRuntime.refresh_models()` rebuilds every alias from the changed sorted catalog. Adding a reference before the retained model changes the alias target; removing a reference before it can make the old alias unresolvable.
   - **Cause**
     - `src/free_claude_code/runtime/application.py:265` invokes `_seed_picker_aliases_from_cache()`, which delegates at line 359 to `seed_picker_aliases()`. That function replaces all global maps (`src/free_claude_code/core/gateway_model_ids.py:99-102`) and numbers refs by `sorted(provider_model_refs)` (line 89), so aliases are positional rather than stable over a running client session.
   - **Fix**
     - Do not rebuild aliases in a way that renumbers existing references during a running process. Preserve existing alias-to-reference assignments and allocate new monotonically increasing aliases only for newly discovered references; retain mappings needed by connected clients, or otherwise introduce a session-safe alias lifecycle. Add a refresh regression test covering insertion and removal before an already-issued alias.

6. General comment 

   <a href="#">P1</a> **Background model discovery leaves picker alias maps stale**

   - **Bug**
     - `ApplicationRuntime.start()` warms the referenced-model cache, starts a non-blocking discovery refresh, then seeds picker aliases only once. When that background refresh adds a model, the catalog includes it but `picker_alias_for()` returns `None`; thus `build_models_list_response()` falls back to canonical `anthropic/...` IDs, which the reported Claude Desktop flow filters out.
   - **Cause**
     - The alias maps are seeded synchronously from the startup cache snapshot at `src/free_claude_code/runtime/application.py:156-157`. `ProviderRuntimeManager.start_model_list_refresh()` launches an unobserved task and no background-refresh completion path calls `_seed_picker_aliases_from_cache()` or otherwise updates the maps.
   - **Fix**
     - Reseed/update picker aliases atomically after background model-list refresh successfully changes the cache, while preserving stable aliases for existing refs; add this regression as a production test.

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into pr1-claude-desk..."](https://github.com/alishahryar1/free-claude-code/commit/6e0c72d0cdaa37fe888c0e2b5a6d92c05328d9b1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48780004)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/.superpowers/sdd/progress.md
+++ b/.superpowers/sdd/progress.md
@@ -1,0 +1,27 @@
+# SDD Progress Ledger
+
+Plan: `docs/superpowers/plans/2026-07-26-integration-fixes.md`
+
+## Task Status
+
+| Task | Status | Branch | Notes |
+|------|--------|--------|-------|
+| 1.1 Add enable_native_tool_approvals Setting | complete | main | 3f13b30 |
+| 1.2 Disable Web Tool Intercept | complete | main | 845f486 |
+| 1.3 Disable Local Optimization Intercept | complete | main | 0f67f90 |
+| 1.4 Integration Test Tool Approval Flow | complete | main | a1967d17f16010c01 |
+| 2.1 Add MCP Optional Dependency | complete | main | 679d9ca |
+| 2.2 Add enable_mcp Setting | complete | main | |
+| 2.3 Create MCPClientManager | complete | main | |
+| 2.4 Add Schema Mapping | complete | main | |
+| 2.5 Security Guardrail Localhost Only | complete | main | |
+| 2.6 Integrate MCP in Provider | pending | - | |
+| 3.1 Add pypdf Optional Dependency | pending | - | |
+| 3.2 Add File Storage Handler | pending | - | |
+| 3.3 Document Text Extraction Fallback | pending | - | |
+| 3.4 Voice Transcription Handler | pending | - | |
+| 3.5 Add voice_provider Setting | pending | - | |
+| 4.1 Run Full CI Suite | pending | - | |
+| 4.2 Manual Integration Test Checklist | pending | - | |
+
+Current Task: **2.6 Integrate MCP in Provider**
--- a/.superpowers/sdd/task-1.1-brief.md
+++ b/.superpowers/sdd/task-1.1-brief.md
@@ -1,0 +1,64 @@
+# Task 1.1: Add `enable_native_tool_approvals` Setting
+
+## Task Summary
+Add a new `enable_native_tool_approvals: bool = True` setting to the FCC Settings class. This feature flag controls whether native Claude Desktop tool approval dialogs are used instead of FCC's auto-intercepts.
+
+## Location in Plan
+Phase 1: Native Tool Approvals (Issue 1) — First task, foundational for Tasks 1.2-1.4
+
+## Files to Modify
+- **Modify:** `src/free_claude_code/config/settings.py`
+- **Test:** `tests/config/test_settings.py`
+
+## Interfaces
+- **Produces:** `Settings.enable_native_tool_approvals: bool` (default `True`)
+- **Consumed by:** Tasks 1.2 and 1.3 (intercept methods in messages.py)
+
+## Exact Test Code
+
+```python
+# tests/config/test_settings.py
+def test_settings_enable_native_tool_approvals_default_true():
+    from free_claude_code.config.settings import Settings
+    s = Settings()
+    assert s.enable_native_tool_approvals is True
+
+def test_settings_enable_native_tool_approvals_configurable():
+    from free_claude_code.config.settings import Settings
+    s = Settings(enable_native_tool_approvals=False)
+    assert s.enable_native_tool_approvals is False
+```
+
+## Exact Implementation Code
+
+```python
+# src/free_claude_code/config/settings.py
+# Add to Settings class:
+enable_native_tool_approvals: bool = True
+```
+
+## Commands to Run
+
+```bash
+# Step 2: Verify test fails
+pytest tests/config/test_settings.py::test_settings_enable_native_tool_approvals_default_true -v
+
+# Step 4: Verify test passes after implementation
+pytest tests/config/test_settings.py::test_settings_enable_native_tool_approvals_default_true -v
+
+# Step 5: Commit
+git add src/free_claude_code/config/settings.py tests/config/test_settings.py
+git commit -m "feat: add enable_native_tool_approvals setting"
+```
+
+## Global Constraints (from plan)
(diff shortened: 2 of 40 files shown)


Gakpey:
Currently, this supports linux only

Alishahryar1:
Desktop launcher is already present

PR #1513, opened 2026-08-22
Add /v1/chat/completions endpoint for OpenAI Chat Completions compatibility

macmaker16:
## Summary

Adds a `/v1/chat/completions` endpoint that accepts OpenAI Chat Completions requests, converts them to Anthropic Messages format, routes through the existing provider execution pipeline, and converts Anthropic SSE responses back to Chat Completions SSE format.

## What changed

**New files:**
- `core/chat_completions/__init__.py` — Module exports
- `core/chat_completions/models.py` — Pydantic models for Chat Completions requests
- `core/chat_completions/converter.py` — OpenAI Chat Completions → Anthropic Messages converter
- `core/chat_completions/stream.py` — Anthropic SSE → Chat Completions SSE stream converter
- `api/handlers/chat_completions.py` — Handler with proper lease management

**Modified:**
- `api/routes.py` — Added `POST /v1/chat/completions` with lease lifecycle

## Why

This enables any OpenAI Chat Completions client (Freebuff, LM Studio, etc.) to use free-claude-code's provider routing and model catalog. The existing `/v1/messages` (Anthropic) and `/v1/responses` (OpenAI Responses) endpoints remain unchanged.

## Testing

- All `ruff check` and `ruff format` checks pass
- All `ty` type checks pass
- Syntax validation passes for all new files

🤖 Generated with Codebuff
Co-Authored-By: Codebuff <noreply@codebuff.com>

<h3>Greptile Summary</h3>

This change adds an OpenAI-compatible Chat Completions endpoint backed by the existing Anthropic Messages execution path. Four blocking behaviors remain: omitted `stream` requests return SSE instead of a JSON completion; non-streaming completions discard successful content and provider terminal errors; `tool_choice: "none"` can still leave provider tool invocation enabled; and malformed tool-call input reaches a server error rather than request validation.

The previous streaming lifetime-management failure is disproved: a real route execution with `stream: true` returned HTTP 200 SSE output and released the request lease successfully.

<h3>Confidence Score: 0/5</h3>

The endpoint is not ready to merge because normal client requests can receive the wrong response format, non-streaming results can be silently lost, tool restrictions can fail open, and malformed input produces server errors.

Four independently reproduced blocking failures remain, including a security-relevant failure to preserve an explicit tool-use prohibition.

**Files Needing Attention:** src/free_claude_code/api/handlers/chat_completions.py, src/free_claude_code/core/chat_completions/converter.py, src/free_claude_code/core/chat_completions/models.py

<details open><summary><h3>Security Review</h3></summary>

The `tool_choice: "none"` conversion is security-relevant because a caller can explicitly prohibit tool use while the provider still receives callable tools with its automatic selection behavior. This can permit tool execution outside the caller's requested authorization boundary.
</details>

**What T-Rex did**
- T-Rex produced a proof for a posted P1 finding, including hermetic reproduction sources and an observed FastAPI TestClient response.
- T-Rex validated the aggregate probe for Chat Completions and captured both the probe source and the runtime output.
- T-Rex executed a focused Chat Completions validation script to exercise tool-choice-none behavior and captured the numbered reproduction script and payloads.
- T-Rex performed a real ASGI route validation for chat completions streaming, confirming 200 responses with SSE data and that lifetime tests passed.
- T-Rex captured regression and lifetime test artifacts for chat completions streaming, including before/after logs and regression test results.

<a href="https://app.greptile.com/trex/runs/20160954/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<h3>Comments Outside Diff (6)</h3>

1. General comment 

   <a href="#">P1</a> **Streaming Chat Completions responses are rejected by the response-lifetime contract**

   - **Bug**
     - `POST /v1/chat/completions` with `stream: true` reaches the mocked provider but fails at the HTTP adapter boundary. The observed client response is HTTP 500 Internal Server Error with `Streaming API responses must use ManagedStreamingResponse.` instead of an SSE stream.
   - **Cause**
     - `src/free_claude_code/api/handlers/chat_completions.py:71-79` returns FastAPI's plain `StreamingResponse`. `src/free_claude_code/api/routes.py:285` unconditionally passes that response to `bind_response_lifetime`, and `src/free_claude_code/api/response_streams.py:158-172` explicitly raises `TypeError` for any `StreamingResponse` that is not `ManagedStreamingResponse`.
   - **Fix**
     - Construct the Chat Completions streaming result as `ManagedStreamingResponse` (or route it through the existing managed first-chunk streaming-response helper) so `bind_response_lifetime` can bind and release the runtime lease.

2. General comment 

   <a href="#">P1</a> **Non-stream chat completions converts the aggregate tuple instead of its message and ignores terminal SSE errors**

   - **Bug**
     - At `src/free_claude_code/api/handlers/chat_completions.py:62-63`, `aggregate_anthropic_sse_to_message(streamed)` returns `(message_dict, error_dict_or_none)`, but the complete tuple is passed to `_to_non_streaming_response`. Since the converter checks attributes such as `.content` and `.stop_reason`, which a tuple lacks, it emits an empty assistant response. A successful mocked provider SSE returned HTTP 200 with `choices[0].message.content = null` rather than the emitted text. A mocked terminal `overloaded_error` likewise returned HTTP 200 with an empty response, losing the provider error message.
   - **Cause**
     - The handler assigns the two-value aggregate result to one variable (`message`) rather than unpacking `message, error`; it also has no branch to translate a non-null terminal SSE error into an OpenAI-compatible error response.
   - **Fix**
     - Unpack the return at line 62 (`message, error = await ...`), pass only `message` to `_to_non_streaming_response`, and translate a non-null `error` to the endpoint's appropriate non-2xx OpenAI error response before constructing a completion.

3. General comment 

   <a href="#">P1</a> **Chat Completions tool_choice none is converted to provider auto**

   - **Bug**
     - A client request declaring a function tool and `tool_choice: "none"` does not suppress the tool. The Chat Completions converter preserves `payload["tools"]` while omitting `payload["tool_choice"]`; the downstream OpenAI-compatible provider request contract interprets that combination as implicit automatic tool use and emits `tool_choice: "auto"`.
   - **Cause**
     - `src/free_claude_code/core/chat_completions/converter.py:150-163` unconditionally serializes request tools before tool-choice processing. At lines `170-172`, the `"none"` branch performs only `pass`, despite the comment claiming no `tool_choice` means no tools. In this stack, absent tool choice with retained tools means auto, not disabled.
   - **Fix**
     - When `request.tool_choice == "none"`, omit/remove `tools` from the converted payload (and omit `tool_choice`), matching the existing OpenAI Responses conversion contract. Add a regression test through the Chat Completions converter and OpenAI-compatible provider request builder asserting that neither provider tools nor an auto tool choice is forwarded.

4. General comment 

   <a href="#">P1</a> **Omitted stream field defaults Chat Completions requests to SSE**

   - **Bug**
     - An OpenAI Chat Completions request without a `stream` member receives an SSE response instead of a JSON `chat.completion` completion envelope.
   - **Cause**
     - `ChatCompletionsHandler.create` checks only `if request_data.stream is False` before choosing aggregation. The Pydantic field defaults to `None`, so an omitted field follows the streaming fallback branch.
   - **Fix**
     - Make the streaming branch explicit (`if request_data.stream is True`) and make omitted/false stream requests use the non-streaming JSON aggregation branch; add a route-level TestClient regression test for omitted `stream`.

5. General comment 

   <a href="#">P1</a> **Chat Completions non-streaming path passes aggregate tuple as a message and suppresses terminal provider errors**

   - **Bug**
     - `aggregate_anthropic_sse_to_message` produced a tuple with arity 2 for both test streams. On success, aggregation produced text `hello from provider`, but `ChatCompletionsHandler.create` returned a 200 completion whose `choices[0].message.content` was null. On the terminal provider-error stream, aggregation returned `{"type":"overloaded_error","message":"provider overloaded"}`, but the handler still returned the same 200 empty completion.
   - **Cause**
     - At `src/free_claude_code/api/handlers/chat_completions.py:65-66`, `message = await aggregate_anthropic_sse_to_message(streamed)` stores the `(message, error)` tuple and passes it to `_to_non_streaming_response`. That converter only reads attribute-style `content` and `stop_reason`, so it cannot convert the tuple (or the aggregated dict). The handler also contains no branch to map the returned terminal SSE error to an HTTP error response.
   - **Fix**
     - Unpack the aggregator result (`message, error = ...`). If `error` is present, return a protocol-appropriate non-2xx error response using the same terminal-error mapping approach as the Messages handler. For success, make `_to_non_streaming_response` consume the aggregate dict (or convert it to the expected message model) and map dict content blocks, stop reason, and usage.

6. General comment 

   <a href="#">P1</a> **Malformed Chat Completions tool calls cause a server error after request validation**

   - **Bug**
     - The public `/v1/chat/completions` request model accepts a string as an element of `messages[].tool_calls`; conversion then calls `.get()` on that string and the handler returns HTTP 500 instead of rejecting the request as invalid.
   - **Cause**
     - `ChatCompletionsMessage.tool_calls` is declared `list[Any] | None`, while `ChatCompletionsToAnthropicConverter.to_anthropic_payload()` treats every element as a dictionary with `function`, `id`, and related fields.
   - **Fix**
     - Replace `list[Any]` with a typed Pydantic tool-call model (including typed function/name/arguments fields), or add an explicit mapping/type validation before conversion; invalid elements should produce FastAPI's 422 validation response rather than a 500.

<sub>Reviews (2): Last reviewed commit: ["fix: use ManagedStreamingResponse for Ch..."](https://github.com/alishahryar1/free-claude-code/commit/10cbad5464a18949b1b115271614b709b8e7e96e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55831273)</sub>

> Greptile also left **4 inline comments** on this PR.

### DIFF
--- a/src/free_claude_code/api/handlers/chat_completions.py
+++ b/src/free_claude_code/api/handlers/chat_completions.py
@@ -1,0 +1,160 @@
+"""OpenAI Chat Completions API product flow."""
+
+import json
+
+from fastapi.responses import JSONResponse, StreamingResponse
+from loguru import logger
+
+from free_claude_code.application.errors import ApplicationError
+from free_claude_code.application.execution import ProviderExecutor
+from free_claude_code.application.routing import ModelRouter
+from free_claude_code.config.settings import Settings
+from free_claude_code.core.anthropic import MessagesRequest
+from free_claude_code.core.chat_completions.converter import (
+    ChatCompletionsToAnthropicConverter,
+)
+from free_claude_code.core.chat_completions.models import ChatCompletionsRequest
+from free_claude_code.core.chat_completions.stream import (
+    chat_completions_sse_from_anthropic,
+)
+from free_claude_code.core.diagnostics import safe_exception_message
+
+
+class ChatCompletionsHandler:
+    """Handle OpenAI Chat Completions requests by converting to Anthropic format.
+
+    The route acquires a RequestRuntimeLease and passes the already-resolved
+    provider_executor here. This handler only owns conversion logic.
+    """
+
+    def __init__(
+        self,
+        settings: Settings,
+        *,
+        model_router: ModelRouter | None = None,
+        provider_executor: ProviderExecutor,
+    ) -> None:
+        self._model_router = model_router or ModelRouter(settings)
+        self._provider_executor = provider_executor
+
+    async def create(self, request_data: ChatCompletionsRequest) -> object:
+        """Create a Chat Completions response."""
+        try:
+            anthropic_payload = (
+                ChatCompletionsToAnthropicConverter.to_anthropic_payload(request_data)
+            )
+            response_request = MessagesRequest(**anthropic_payload)
+            routed = self._model_router.resolve_messages_request(response_request)
+
+            streamed = self._provider_executor.stream(
+                routed,
+                wire_api="messages",
+                raw_log_label="CHAT_COMPLETIONS_PAYLOAD",
+                raw_log_payload=request_data.model_dump(mode="json", exclude_none=True),
+                request_id="",
+            )
+
+            if request_data.stream is False:
+                from free_claude_code.core.anthropic import (
+                    aggregate_anthropic_sse_to_message,
+                )
+
+                message = await aggregate_anthropic_sse_to_message(streamed)
+                return self._to_non_streaming_response(message, request_data.model)
+
+            async def event_stream():
+                async for chunk in chat_completions_sse_from_anthropic(
+                    streamed, request_data.model
+                ):
+                    yield chunk
+
+            return StreamingResponse(
+                event_stream(),
+                media_type="text/event-stream",
+                headers={
+                    "Cache-Control": "no-cache",
+                    "Connection": "keep-alive",
+                    "X-Accel-Buffering": "no",
+                },
+            )
+
+        except ApplicationError:
+            raise
+        except Exception as exc:
+            logger.error("Chat Completions error: {}", safe_exception_message(exc))
+            return JSONResponse(
+                status_code=500,
+                content={
+                    "error": {
+                        "message": safe_exception_message(exc),
+                        "type": "server_error",
+                    }
+                },
+            )
+
+    def _to_non_streaming_response(self, message: object, model: str) -> dict:
+        """Convert an aggregated Anthropic message to Chat Completions format."""
+        content = ""
+        tool_calls: list[dict] = []
+        finish_reason = "stop"
+
+        if hasattr(message, "content"):
+            blocks = (
+                message.content
+                if isinstance(message.content, list)
+                else [message.content]
+            )
+            for block in blocks:
+                if not hasattr(block, "type"):
+                    continue
+                if block.type == "text":
+                    content += block.text
+                elif block.type == "tool_use":
+                    tool_calls.append(
+                        {
+                            "id": block.id,
+                            "type": "function",
+                            "function": {
+                                "name": block.name,
+                                "arguments": json.dumps(block.input)
+                                if isinstance(block.input, dict)
+                                else str(block.input),
+                            },
+                        }
+                    )
+
+        if hasattr(message, "stop_reason"):
+            stop_map = {
+                "end_turn": "stop",
+                "stop_sequence": "stop",
+                "max_tokens": "length",
+                "tool_use": "tool_calls",
+            }
+            finish_reason = stop_map.get(message.stop_reason, "stop")
+
+        choice: dict = {
+            "index": 0,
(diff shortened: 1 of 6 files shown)


Alishahryar1:
What is this needed for? None of the providers use it and it creates an extra surface for no reason.

PR #1527, opened 2026-08-23
Cut truncated transcripts on safe markup boundaries

SEPURI-SAI-KRISHNA:
## Problem

When a transcript update exceeds the platform limit, `render_segments` drops whole segments and then keeps a **tail of the last one**. That last segment is already-rendered markup, but the tail was taken with a raw slice and a raw ellipsis:

```python
tail = "..." + last_part[-(budget - 3):] if len(last_part) > budget else last_part
```

Two independent defects, both producing MarkdownV2 that Telegram rejects with `400 Bad Request: can't parse entities`:

1. **The `"..."` is never escaped.** `prefix_marker` is built with `ctx.escape_text(...)` and comes out as `\.\.\.`, but this second ellipsis is spliced in raw. Three bare `.` characters are enough on their own.
2. **The slice lands anywhere.** `last_part` is finished markup, so an arbitrary offset can split a `\.` escape pair (stranding the backslash and unescaping the character), cut into a `` ``` `` block opened by a `ThinkingSegment` or `ToolResultSegment`, or leave a `*`/`_`/`~` entity unclosed.

Sweeping `limit_chars` from 120 to 400 over one oversized plain-text segment: **280/280 outputs were invalid MarkdownV2**, and **38 of them also broke a backslash escape pair**.

The update is not lost — `telegram_io._with_retry` catches the entity error and resends with `parse_mode=None`. But that costs an extra round-trip on every truncated update and delivers it as raw text, so the user sees `\.\.\. \(truncated\)` and `step\-74\.done` with every escape backslash exposed and all formatting gone.

## Why the fix is not local to `renderer.py`

`render_segments` is deliberately platform-agnostic — it only has the `RenderCtx` callables and knows nothing about MarkdownV2 or Discord syntax. Deciding where finished markup may be cut *requires* that syntax. Putting the rules in the renderer would hard-code Telegram into a shared component; leaving them out means it cannot cut safely at all.

So the cut point moves to the layer that owns the syntax, matching how `bold`, `escape_code` and `render_markdown` already work.

## Changes

| Before | After |
| --- | --- |
| `renderer.py` sliced rendered markup at a raw character offset. | `renderer.py` asks `ctx.tail_slice(...)` and falls back to dropping the segment when no safe cut exists. |
| A second, unescaped `"..."` was spliced into the markup. | Removed — `prefix_marker` already announces the truncation, and it is escaped. |
| No platform owned truncation rules. | New `messaging/rendering/tail_slice.py` finds standalone cut points; `mdv2_tail_slice` and `discord_tail_slice` supply each platform's paired delimiters. |
| `RenderCtx` had no truncation hook. | `RenderCtx.tail_slice` defaults to `keep_whole_or_drop`, which never cuts inside markup, so an unwired context degrades safely instead of emitting invalid output. |
| Package version was 5.13.10. | Package version is 5.13.11. |

A cut is offered only where the remainder stands alone: no escape pair straddles it, and it is not inside a code span, fenced block, link, or open inline entity. If the markup is already malformed (an unterminated fence), no cut point is offered at all and the segment is dropped rather than sliced.

## Content is preserved, not sacrificed

Validity would be trivial to get by dropping everything, so retention was measured against the budget:

| Rendered segment | Retained |
| --- | --- |
| Escaped prose (the realistic oversized case) | **99–100%** of budget |
| Text with inline `*bold*` / `` `code` `` entities | **99%** of budget |
| A single unbroken fenced block | ~0% — no interior cut point exists |

The last row is the honest trade-off: you cannot cut into a fenced block without orphaning its opening fence, so such a segment degrades to the truncation marker. That is a deliberate choice — a short, correctly formatted message reads better than a longer one that loses all formatting to the plain-text fallback. In practice segments self-truncate first (`thinking_tail_max=1000`, `tool_output_tail_max=1600`), so the oversized segment reaching this path is almost always escaped prose, which retains ~100%.

## Verification

The original sweep is now **0/280 invalid**, with **no marker-only fallbacks**, across three shapes: plain escaped text, a segment containing a fenced block plus bold, and truncation with a status suffix. Outputs were checked by walking them the way Telegram's parser does — honouring backslash escapes, code spans and fenced blocks — asserting no reserved character survives unescaped outside an entity.

16 new tests cover the slicer directly (escape pairs, fences, inline entities, links, longest-valid-suffix, malformed input, Discord's two-character delimiters) and the renderer end to end through the real Telegram profile.

`_ctx()` in `tests/messaging/test_transcript.py` mirrors the Telegram profile field for field and now wires `tail_slice=mdv2_tail_slice`, so `test_transcript_truncation_preserves_last_segment_tail` continues to exercise real tail retention rather than the conservative default.

`./scripts/ci.sh --skip playwright` passes: ruff format, ruff check, ty, and 3669 tests (114 skipped). Playwright was skipped locally as it only drives the Admin UI, which this change does not touch; CI will run it.

## Note on ordering

Branched from `main` at 5.13.10 (after #1526 merged) and takes 5.13.11.

<h3>Greptile Summary</h3>

Transcript-tail rendering now keeps Discord links with balanced parentheses intact when a message must be truncated, preventing stray closing delimiters from appearing in delivered text.

No blocking failure remains.

The exercised Discord link-truncation path preserves standalone markup for every truncated character budget.

**What T-Rex did**
- I ran the discord-tail-link regression script against 58 budgets and confirmed the current behavior produced zero violations while the legacy approach produced 34 violations of trailing content.
- I traced the current behavior to code paths in discord\_markdown.py and tail\_slice.py, where balanced\_link\_parens is passed to safe\_tail, nested destination parentheses are tracked, and a standalone suffix is selected.
- I reviewed the regression artifacts, including the executable regression script and two logs, to verify the comparison between legacy and current results.

<a href="https://app.greptile.com/trex/runs/20333533/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(messaging): match nested link parens..."](https://github.com/alishahryar1/free-claude-code/commit/4860690087d30f4623d244b439b1657cea376eb4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55983482)</sub>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "free-claude-code"
-version = "5.14.2"
+version = "5.14.3"
 description = "Local proxy connecting coding agents to OpenAI-compatible AI providers"
 readme = "README.md"
 requires-python = ">=3.14.0"
--- a/src/free_claude_code/messaging/rendering/discord_markdown.py
+++ b/src/free_claude_code/messaging/rendering/discord_markdown.py
@@ -31,7 +31,11 @@ def escape_discord_code(text: str) -> str:
 
 def discord_tail_slice(text: str, max_chars: int) -> str:
     """Return a valid Discord-markdown suffix of rendered text within ``max_chars``."""
-    return safe_tail(text, max_chars, DISCORD_INLINE_DELIMITERS)
+    # Discord escapes neither parenthesis in a link destination, so nesting
+    # must be matched to find the real end of the link.
+    return safe_tail(
+        text, max_chars, DISCORD_INLINE_DELIMITERS, balanced_link_parens=True
+    )
 
 
 def discord_bold(text: str) -> str:
--- a/src/free_claude_code/messaging/rendering/discord_markdown.py
+++ b/src/free_claude_code/messaging/rendering/discord_markdown.py
@@ -7,9 +7,12 @@ Used by the message handler and Discord platform adapter.
 from markdown_it import MarkdownIt
 
 from .markdown_tables import normalize_gfm_tables
+from .tail_slice import safe_tail
 
 # Discord escapes: \ * _ ` ~ | >
 DISCORD_SPECIAL = set("\\*_`~|>")
+# Paired inline markers this renderer emits, longest first.
+DISCORD_INLINE_DELIMITERS = ("**", "__", "~~", "||", "*", "_")
 
 _MD = MarkdownIt("commonmark", {"html": False, "breaks": False})
 _MD.enable("strikethrough")
@@ -24,6 +27,11 @@ def escape_discord(text: str) -> str:
 def escape_discord_code(text: str) -> str:
     """Escape text for Discord code spans/blocks."""
     return text.replace("\\", "\\\\").replace("`", "\\`")
+
+
+def discord_tail_slice(text: str, max_chars: int) -> str:
+    """Return a valid Discord-markdown suffix of rendered text within ``max_chars``."""
+    return safe_tail(text, max_chars, DISCORD_INLINE_DELIMITERS)
 
 
 def discord_bold(text: str) -> str:
--- a/src/free_claude_code/messaging/rendering/profiles.py
+++ b/src/free_claude_code/messaging/rendering/profiles.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 from free_claude_code.messaging.rendering.discord_markdown import (
     discord_bold,
     discord_code_inline,
+    discord_tail_slice,
     escape_discord,
     escape_discord_code,
     render_markdown_to_discord,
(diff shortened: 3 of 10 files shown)


SEPURI-SAI-KRISHNA:
Rebased onto `7fedb42` and force-pushed. That replaces the rebase you pushed as `efd2e61`, so, why:

Your rebase landed on `73dfbd23`, but #1536, #1540 and #1541 merged a few minutes later and moved `main` to `5.14.2`. That reintroduced the `pyproject.toml` / `uv.lock` conflict this PR was already blocked on. So it's rebased onto current
`main` and bumped `5.13.11` → `5.14.3` (patch, still a bug fix), with `uv lock` regenerated in the same commit.

No source changes: all eight source and test files are byte-identical to `efd2e61`. Only the base and the version line differ.

Also re-ran the suite against #1541's streaming refactor, since your base predated it, 3702 passed, 114 skipped locally, no interaction with the `messaging/` changes here.

PR #1544, opened 2026-08-24
feat: enable fcc-desktop on Linux with console fallback

Gakpey:
## Summary

`fcc-desktop` currently hard-rejects every platform except Windows and macOS (`desktop_entrypoint.py`), so Linux users get:

```
FCC Desktop is supported on Windows and macOS.
```

This PR enables the desktop shell on Linux with feature parity plus graceful degradation.

## What changes

The controller layer (`cli/desktop.py`) is already platform-neutral — no changes there. This PR only touches the edges that treat Linux as unsupported:

| File | Change |
|---|---|
| `desktop_entrypoint.py` | Platform gate now accepts `linux`; linux routes through a tray-availability probe |
| `desktop_tray.py` | New `tray_is_available()` probe: constructs a pystray `Icon` (which resolves the backend module) and discards it; on failure returns an actionable reason (AppIndicator package names / X11 session hint) |
| `desktop_console.py` (new) | Null tray: `run()` blocks on an Event, `stop()` sets it. The unchanged `DesktopController.run()` lifecycle still owns server start/stop, singleton lock, and clean shutdown. SIGINT propagates out of `Event.wait()` into the existing cleanup chain |
| `pyproject.toml` | `pystray`/`Pillow` markers extended to all platforms; missing system tray libraries are handled at runtime by the probe, not at install time |
| `scripts/install.sh` / `uninstall.sh` | Linux writes/removes an owned XDG `.desktop` entry + exported PNG icon under `~/.local/share`, with the same owned-file guard pattern as the macOS bundle logic |

## Behavior

- **Tray backend available** (AppIndicator/GTK/X11): identical tray experience to macOS/Windows — Open Admin / Check Status / Restart / Quit
- **No tray backend** (e.g. GNOME Wayland without the AppIndicator extension): prints the reason plus install hints, then runs in **console mode** — same singleton lock, foreground server, clean Ctrl+C shutdown
- Other platforms keep the existing rejection message

## Testing

- 12 new unit tests: probe success/import-failure/artwork-failure paths, null-tray blocking semantics, full controller lifecycle through the null tray, entrypoint routing table per platform
- Installer/uninstaller script tests for the Linux `.desktop` entry: creation, icon export, non-FCC entry preservation, failure path, uninstall removal of FCC-owned entries only
- Live smoke on a real Linux desktop: probe returns available, `--export-icon` works

Version bumped to 5.15.0 (MINOR: new platform capability).

<h3>Greptile Summary</h3>

This change adds Linux desktop launching, Claude Desktop routing, and local TLS proxy support. Four user-facing failures remain: trayless Linux launches cannot be stopped from the application menu, uninstall can delete an unrelated icon, launcher paths containing spaces or apostrophes cannot start, and Claude Desktop cleanup removes a pre-existing discovery preference.

<h3>Confidence Score: 1/5</h3>

Not safe to merge until the desktop launcher lifecycle, ownership cleanup, path quoting, and Claude Desktop preference restoration are corrected.

Four independent user-facing failures remain in the Linux desktop and Claude Desktop flows.

**Files Needing Attention:** src/free_claude_code/cli/desktop_console.py, scripts/install.sh, scripts/uninstall.sh, src/free_claude_code/cli/claude_desktop.py

**What T-Rex did**
- T-Rex ran the trayless Linux launcher probe and captured before/after traces and the Linux desktop regression test output.
- T-Rex isolated and documented the Linux icon-uninstall probe and its output.
- T-Rex isolated Claude Desktop discovery-setting probe and captured its output.
- T-Rex validated Desktop Entry Exec handling by comparing single-quoted versus double-quoted paths, confirming that double-quoted resolves the special-character path.
- T-Rex performed general contract validations across lifecycle, icon deletion, and launcher execution scenarios, summarizing observed outputs and source probes.

<a href="https://app.greptile.com/trex/runs/20418133/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<h3>Comments Outside Diff (4)</h3>

1. General comment 

   <a href="#">P1</a> **Trayless app-launcher fallback has no user-accessible shutdown path**

   - **Bug**
     - If native tray probing fails, launching the installed Linux desktop entry starts the console fallback with no terminal attached. The fallback starts the server and waits indefinitely, while no tray, terminal, or UI action is available to invoke its stop method. The user must externally signal/kill the process to stop it.
   - **Cause**
     - `ConsoleDesktopTray.run()` waits on `_stop_event`, but the only setter is its programmatic `stop()` method. The installed desktop entry explicitly specifies `Terminal=false`, so the documented SIGINT path is unavailable to an app-launcher user.
   - **Fix**
     - Provide an accessible lifecycle control in trayless mode, such as a visible terminal (`Terminal=true`) with documented Ctrl+C, a GUI/dialog shutdown action, or an authenticated local Admin/API stop control with a documented launcher/UI entry.

2. General comment 

   <a href="#">P1</a> **Uninstall deletes an unowned regular Linux icon**

   - **Bug**
     - `remove_linux_desktop_entry` removes `$XDG_DATA_HOME/icons/free-claude-code.png` whenever it is a non-symlink regular file. The executed isolated-HOME/XDG probe confirmed deletion when the desktop entry was absent and when it existed but was explicitly unowned.
   - **Cause**
     - The icon deletion branch at `scripts/uninstall.sh:263-264` checks only whether the icon is a regular non-symlink file; it has no ownership marker or dependency on an owned desktop entry.
   - **Fix**
     - Only remove the icon when ownership can be established, such as by tying it to a verified owned desktop entry or a dedicated ownership marker.

3. General comment 

   <a href="#">P1</a> **Unconfigure removes a discovery setting that existed before FCC configuration**

   - **Bug**
     - When a Claude Desktop config already has `modelDiscoveryEnabled: true`, running configure and then unconfigure removes that setting rather than restoring the pre-existing state.
   - **Cause**
     - `unconfigure_claude_desktop_config` checks only whether `_DISCOVERY_KEY` exists and deletes it unconditionally; it does not record whether FCC added the key or distinguish a user-owned `true` value.
   - **Fix**
     - Track FCC ownership of the discovery setting (or preserve and restore its prior value) and only remove `modelDiscoveryEnabled` when configure added it.

4. General comment 

   <a href="#">P1</a> **Linux desktop launcher misquotes tool-bin paths with spaces or apostrophes**

   - **Bug**
     - At `scripts/install.sh:1264`, the generated `.desktop` entry uses `shell_quote`, producing a single-quoted shell token. Desktop Entry Exec parsing does not treat single quotes as grouping characters. For a path with spaces and an apostrophe, the command is split and cannot launch the intended `fcc-desktop` executable.
   - **Cause**
     - `shell_quote` at `scripts/install.sh:1137-1140` implements POSIX shell quoting (`'...'\''...'`), but Desktop Entry `Exec` requires Desktop Entry quoting rules, where double quotes—not single quotes—group whitespace-containing arguments.
   - **Fix**
     - Generate the Exec command with Desktop Entry-compatible double-quote/backslash escaping (or use an executable path without special characters), rather than reusing `shell_quote`.

<sub>Reviews (4): Last reviewed commit: ["test: skip no-node dsh scenario when sys..."](https://github.com/alishahryar1/free-claude-code/commit/bf0632542dd0bf586148d4567d7a16ce4040f008) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56264714)</sub>

> Greptile also left **4 inline comments** on this PR.


Gakpey:
Superseded by #1545, which contains this change set plus the desktop-scoped picker aliasing and x-api-key auth fix.

Issue #1043, opened 2026-07-11
Windows: fcc-claude launches Claude but does not route requests through fcc-server


DaveWZz:
OS: Windows
Installed from latest install.ps1, ran fcc-server and started successfully, as NVIDIA provider validates. fcc-claude launched Claude Code, and authenticated successfully. But requests never reached fcc-server (/v1/messages never appears in logs), and Claude uses Anthropic API instead. curl http://127.0.0.1:8082/v1/models reaches the proxy (401 Missing API key), confirming the server is alive.

Intelli2Byte:
Hi! @Alishahryar1  I'd like to work on this issue. I'll investigate why fcc-claude doesn't 
properly route requests through fcc-server on Windows. 

Plan:
1. Investigate the fcc-claude launcher script and how it sets ANTHROPIC_BASE_URL
2. Check if environment variables are properly passed to the subprocess on Windows
3. Fix the routing issue and add tests

Is anyone else already working on this? If not, I'll get started.

Fardin7798:
Opened #1621 with a possible root cause: the env var filter that strips inherited `ANTHROPIC_*` vars before setting the proxy's own `ANTHROPIC_BASE_URL` does the check case-sensitively, but Windows env vars are case-insensitive - so a differently-cased leftover (e.g. `Anthropic_Base_Url`) could survive the filter and create ambiguity about which value Claude Code actually reads. I don't have a Windows machine to confirm this reproduces your exact setup, but if you get a chance to check `echo %ANTHROPIC_BASE_URL%` / look for any Anthropic-prefixed system or user environment variables in Windows System Properties before running `fcc-claude`, that would help confirm or rule it out.

Intelli2Byte:
Tested PR #1621 on Windows 11 — confirmed the fix works. 
Requests now route through fcc-server even with mixed-case 
ANTHROPIC_BASE_URL set. See my comment on #1621 for details.

PR #1621, opened 2026-08-30
fix(claude): filter inherited ANTHROPIC_* env vars case-insensitively

Fardin7798:
Relates to #1043.

## What
`build_claude_proxy_env()` strips inherited `ANTHROPIC_*` env vars before setting `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` itself, but did the strip with a case-sensitive `key.startswith("ANTHROPIC_")` check (same gap existed for the `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` strip).

## Why this looks like the #1043 root cause
Windows environment variables are case-insensitive, so a leftover differently-cased entry - e.g. `Anthropic_Base_Url`, which can end up set by an installer or added manually via the System Properties GUI (Windows users don't get the same all-caps shell convention Unix users do) - survives a case-sensitive filter and sits in the same env dict alongside the value FCC is trying to set. The two keys are distinct to this Python dict, but the same variable to Windows itself and to a child process runtime that resolves env lookups case-insensitively, so which one Claude Code actually reads becomes ambiguous. That matches the reported symptom exactly: the proxy launches, Claude authenticates successfully, but `/v1/messages` never reaches `fcc-server` and traffic goes straight to the real Anthropic API instead.

I don't have a Windows machine to reproduce the original report directly, so I want to be upfront: this is the most concrete, evidence-based cause I could find by reading the code carefully (case-sensitive filtering + Windows' case-insensitive env vars is a well-known class of bug), not a confirmed-on-hardware fix. It's a real correctness bug either way - the filter clearly does not do what its own comment says on Windows - so I think it's worth shipping regardless, but @Alishahryar1 or the issue reporter may be able to confirm whether it was in fact the cause.

## Fix
Filter by casefolded key name instead, which removes the ambiguity on every platform (not just Windows) and is a one-line-per-check change.

## Testing
- Added `tests/cli/test_claude_env.py`: same-case behavior (existing coverage gap) + a differently-cased regression case reproducing the #1043 scenario.
- `uv run pytest tests/cli` - 15/15 pass
- `uv run ruff format --check` / `uv run ruff check` - clean
- `uv run ty check` - clean

<h3>Greptile Summary</h3>

This change makes Claude proxy environment sanitization case-insensitive so that inherited mixed-case Anthropic settings cannot coexist with FCC-managed routing on Windows.

The environment builder was exercised with mixed-case inherited Anthropic endpoint, credential, model, and traffic-disable settings. It removed every conflicting case variant, installed only the canonical FCC proxy URL and auth token, and preserved unrelated environment variables. The focused Claude environment tests also passed.

Disproved drop: a mixed-case inherited Anthropic setting surviving into the child environment and overriding FCC routing does not occur in the current implementation.

Safe to merge: the changed environment-construction path removes conflicting case variants before installing FCC-managed proxy settings.

A deterministic executable check exercised the Windows-style case-insensitive failure path and observed only canonical managed values in the resulting environment; the focused test suite passed as well.

**Files Needing Attention:** No files need further attention.

**What T-Rex did**
- Ran the deterministic mixed-case environment validation for phase before and phase after to compare legacy behavior against the current FCC environment builder.
- Executed pytest tests for claude\_env to verify environment handling; all 3 focused tests passed.
- Reviewed the before-capture baseline to understand why the legacy case-sensitive filter was unsafe for Windows-style lookups due to two ANTHROPIC\_BASE\_URL aliases and stale values.
- Reviewed the after-capture results showing the utility removes all case-variant inherited Anthropic entries and the mixed-case traffic flag, then adds only the managed credentials and URL.

<a href="https://app.greptile.com/trex/runs/21481831/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(claude): filter inherited ANTHROPIC\_..."](https://github.com/alishahryar1/free-claude-code/commit/7bee15ca056e848894d3bc8f8b6e6cc176bb2389) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58386781)</sub>

### DIFF
--- a/src/free_claude_code/cli/claude_env.py
+++ b/src/free_claude_code/cli/claude_env.py
@@ -7,6 +7,12 @@ from free_claude_code.cli.local_http import with_local_proxy_bypass
 CLAUDE_CODE_AUTO_COMPACT_WINDOW = "190000"
 CLAUDE_BINARY_NAME = "claude"
 
+# Environment variable names FCC always sets itself; any inherited entry with
+# one of these names is dropped before the block below re-sets it, regardless
+# of casing.
+_ANTHROPIC_PREFIX = "anthropic_"
+_STRIPPED_KEYS = frozenset({"claude_code_disable_nonessential_traffic"})
+
 
 def build_claude_proxy_env(
     *,
@@ -16,13 +22,24 @@ def build_claude_proxy_env(
 ) -> dict[str, str]:
     """Return the canonical environment for Claude Code proxy sessions."""
 
+    # Windows environment variables are case-insensitive, so a leftover
+    # differently-cased entry (e.g. `Anthropic_Base_Url`, set by an installer
+    # or a Windows System Properties GUI) would survive a case-sensitive
+    # startswith() filter and sit alongside the value we set below. The two
+    # keys look distinct to this Python dict, but the child process's runtime
+    # resolves env lookups case-insensitively on Windows, so which one wins
+    # becomes ambiguous - in practice this is how Claude Code has been
+    # observed silently falling back to the real Anthropic API instead of
+    # the local proxy. Filtering by casefolded name avoids the ambiguity
+    # entirely, on every platform.
+    #
     # Claude's aggregate traffic flag also suppresses gateway model discovery.
     env = with_local_proxy_bypass(
         {
             key: value
             for key, value in base_env.items()
-            if not key.startswith("ANTHROPIC_")
-            and key != "CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC"
+            if not key.casefold().startswith(_ANTHROPIC_PREFIX)
+            and key.casefold() not in _STRIPPED_KEYS
         },
         proxy_root_url=proxy_root_url,
     )
--- a/tests/cli/test_claude_env.py
+++ b/tests/cli/test_claude_env.py
@@ -1,0 +1,69 @@
+"""build_claude_proxy_env must always win over inherited Anthropic env vars,
+including on Windows where environment variable names are case-insensitive.
+"""
+
+from free_claude_code.cli.claude_env import build_claude_proxy_env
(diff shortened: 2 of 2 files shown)


Intelli2Byte:
I tested this on Windows 11, PowerShell, Python 3.14.0, FCC 5.17.5.

## Reproduction (before fix)
Set `$env:Anthropic_Base_Url = "https://api.anthropic.com"` (mixed case),
started fcc-server, ran fcc-claude.

Result: Claude Code authenticated but requests went directly to Anthropic.
No `POST /v1/messages` appeared in fcc-server logs — confirming the bug.

## After applying this PR
Same setup with mixed-case env var.

Result: ✅ Requests now route through fcc-server correctly.
Server logs show:

The `POST /v1/messages` reaching fcc-server confirms Claude Code is no longer
bypassing the proxy. The 503 is expected — I don't have a provider API key
configured, so FCC can't forward the request upstream. The routing issue
itself is fixed.

## Normal flow (no mixed-case env var)
✅ Also routes through fcc-server correctly — no regression.

## Tests
- `uv run pytest tests/cli -v` — all 15 pass
- `uv run pytest` — all pass 

## Verdict
The case-insensitive filtering fix works on Windows. LGTM 👍

One question: should we also casefold the `CLAUDE_CODE_*` env var filter?
Windows users could have those set with mixed case too.

Alishahryar1:
Not needed anymore

PR #1694, opened 2026-09-06
Add xKiro provider (OpenAI-compatible multi-vendor gateway)

Rushifakami:
Adds [xKiro](https://xkiro.com) as an OpenAI-chat provider, following the provider-addition pattern of #1670.

## Why

xKiro is a free OpenAI-compatible multi-vendor Chat Completions gateway at `https://api.xkiro.com/v1` (docs: docs.xkiro.com). Its catalog currently lists 112 chat models with vendor-prefixed `vendor/model` ids — the bare model name returns 404 — and a free tier covering 40 of them, which makes it useful for FCC harnesses without a paid subscription.

## How

- Register `xkiro` through the standard provider catalog: descriptor, Settings fields (`XKIRO_API_KEY` / `XKIRO_BASE_URL` / `XKIRO_PROXY`), OpenAI-chat profile, admin manifest entries, `.env.example`, README table row, and a smoke default model taken from the live free tier (`xkiro/deepseek/deepseek-v4-flash`).
- The reasoning profile follows the gateway's documented semantics (docs.xkiro.com/guides/reasoning): omitting `reasoning_effort` is *not* disabling — most models default to reasoning on and bill reasoning tokens — so `NamedEffortReasoning` forwards the client's resolved intent, and the gateway adjusts unsupported levels downward per model instead of rejecting them. `reasoning_content` deltas arrive on the Chat Completions stream and are replayed as thinking history (`ReasoningReplayMode.REASONING_CONTENT`).
- Credential validation intentionally has no probe: `GET /v1/models` is public, so it cannot distinguish a valid key from an invalid one.
- Dedicated provider test file (`tests/providers/test_xkiro.py`) plus the contract/test-runtime synchronization every provider addition requires: catalog order, instantiation table, and the credential-validation no-probe count (20 → 21).
- Version bump (minor, `uv lock` in the same commit), per the provider-addition precedent (#1670).

## Validation

Local `scripts/ci.ps1` (suppressions, ruff format/check, ty, pytest), rebased on latest main (#1782–#1785): 5,486 passed, 144 skipped, 0 failed. Live gateway checks against `api.xkiro.com/v1` confirmed the catalog shape, `vendor/model` routing, `reasoning_effort` acceptance (explicit `none` honored), and streaming deltas. The test file was adapted to the transport refactor in #1783 (`_chat._build_request_body`) and now documents the `REASONING_DEFAULT` semantics inline: omission means "inherit the gateway default", never "reasoning disabled" (per review feedback).

<h2><a href="https://app.greptile.com/api/retrigger?id=61023932"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

Safe to merge; no outstanding issues require changes.

<h3>Summary</h3>

This PR adds xKiro as an OpenAI-compatible multi-vendor gateway, including configuration, Admin metadata, runtime construction, model discovery, credential validation, reasoning translation, documentation, smoke defaults, and focused provider tests.

- Registers the xKiro credential, endpoint, proxy, profile, icon, and default model across the provider catalog.
- Maps client reasoning intent to xKiro request fields and replays streamed reasoning content.
- Adds model-capability parsing and comprehensive provider, runtime, catalog-order, and credential-validation coverage.
- Updates package metadata for the provider addition.

<sub>Reviews (15) · Last reviewed commit: ["Merge remote-tracking branch &#39;xkiro-fork..."](https://github.com/alishahryar1/free-claude-code/commit/79319a0be5e4bf8e75d22e2ebb2dae8c93dc358c)</sub>


aravinditte (approved):
Overall: Approve — the implementation is well aligned with the existing provider architecture and appears safe to merge.

The xKiro integration is wired through the expected layers: provider catalog/descriptor, settings and environment variables, OpenAI Chat profile, runtime registration, admin metadata, documentation, smoke defaults, and dedicated provider tests. The provider descriptor correctly uses XKIRO_API_KEY, https://api.xkiro.com/v1, and the optional proxy configuration.
The reasoning implementation is also thoughtfully handled. In particular, using ReasoningReplayMode.REASONING_CONTENT and NamedEffortReasoning matches the stated xKiro gateway behavior rather than assuming OpenAI-style omission means reasoning is disabled.

The dedicated test suite covers the important provider-specific behavior, including provider construction, reasoning intent mapping, reasoning-content replay through tool history, vendor-prefixed model IDs, and model catalog parsing. That is a strong set of regression tests for this type of integration.

Minor concern

I would add one explicit regression test for the default reasoning behavior.

The current test expects the default FCC reasoning state to result in no reasoning_effort field:

(REASONING_DEFAULT, None)

and then:

assert body.get("reasoning_effort") == expected_effort

However, the implementation comment explicitly says that omitting reasoning_effort is not equivalent to disabling reasoning on xKiro, since many models default to reasoning being enabled.

That makes the default path worth pinning down with a test that documents the intended product behavior: whether FCC's REASONING_DEFAULT is deliberately allowed to inherit xKiro's server-side default, or whether FCC should always send an explicit effort for cost/control consistency.

Everything else looks consistent with the established provider pattern, including the catalog ordering, runtime registration, credential-validation bookkeeping, environment example, README entry, and version bump.
Recommendation: Approve, with the default-reasoning behavior either covered by an explicit regression test or documented as an intentional design choice.

aravinditte (approved):
One thing I would still consider

The only area I'd recommend documenting a little more explicitly is the interaction between FCC's default reasoning policy and xKiro's server-side reasoning default.

The implementation intentionally sends no reasoning_effort for the default policy, while the comments state that omission on xKiro may result in reasoning being enabled by the gateway.

That may be exactly what is intended, but it is important enough that a future maintainer could otherwise interpret the None value as meaning "reasoning disabled." A short comment or a dedicated regression assertion describing the intended semantics would make this behavior much clearer.

Final assessment

I don't see a blocking issue in the submitted changes. The provider is integrated across the relevant architectural layers, the xKiro-specific reasoning and model-ID semantics are accounted for, and the tests cover the important request/response transformations.

Recommendation: Approve / Safe to merge.

The only follow-up I would suggest is making the default reasoning semantics slightly more explicit so that the current behavior remains intentional and easy to maintain.

Rushifakami:
Rebased on latest main (bf59598) and adapted the provider test to the #1783 transport refactor (`_chat._build_request_body`). Also made the `REASONING_DEFAULT` semantics explicit in the test per @aravinditte's review note: the parametrized case and docstring now state that omission means "inherit the gateway default", never "reasoning disabled". CI: 5,486 passed / 144 skipped / 0 failed.

aravinditte (approved):
Review

I reviewed the full diff again. The implementation looks complete and consistent with the existing provider architecture.

What looks good
The provider is registered consistently across the provider catalog, settings, admin manifest, runtime, and contract tests. The XKIRO_API_KEY, XKIRO_BASE_URL, and XKIRO_PROXY configuration is wired through the standard provider mechanisms. - Reusing OpenAIChatProvider is the right approach for an OpenAI-compatible gateway and keeps the change appropriately scoped.
The xKiro reasoning behavior is handled explicitly with ReasoningReplayMode.REASONING_CONTENT and NamedEffortReasoning.
The updated reasoning tests clearly distinguish between REASONING_DEFAULT, explicit reasoning on, and explicit reasoning off. In particular, the comment makes it clear that None means "inherit the gateway default", not "disable reasoning". This resolves the main ambiguity I would have flagged in an earlier review.
The tool-history test covers the important conversion of thinking content into reasoning_content, preservation of the assistant text, tool-call serialization, and subsequent tool results.
Vendor-prefixed model IDs are tested explicitly and passed through unchanged, which is important for xKiro's vendor/model naming scheme.
Model discovery is covered with both paid and free models, so the integration is testing the actual catalog shape rather than only provider construction.
The provider runtime and credential-validation contract updates are included, avoiding the common case where a new provider works locally but is missing from one of the central registries. - .env.example, README, and the smoke configuration are updated, so the provider is discoverable and usable without requiring users to infer the configuration manually. - The dedicated xKiro test file provides good coverage for a provider addition and gives confidence beyond simple registration tests.
Minor suggestion

I don't see a blocking issue in the current diff.

One optional addition would be a test for the case where a requested reasoning level is unsupported by a specific xKiro model and therefore gets downgraded. The current tests verify that FCC sends the intended reasoning_effort, but they do not directly exercise the gateway-side fallback behavior described in the comments. This is not a blocker if that behavior is guaranteed by xKiro itself.

Verdict

Approve

The provider is integrated consistently, the xKiro-specific reasoning and model-ID semantics are covered, and the relevant configuration and contract tests have been updated. I don't see a change that warrants blocking the merge.

Rushifakami:
@Alishahryar1

Rushifakami:
﻿## Update: model metadata, reasoning intent, and key verification (cc3b41e7)

This commit addresses the two confirmed provider-scoped gaps plus key verification, keeping everything scoped to xKiro (no shared FCC behavior changes).

### 1. Model catalog metadata was being dropped
The profile previously left `model_listing` at defaults, so xKiro's documented `/v1/models` metadata was discarded: vision models showed text-only, real context windows (e.g. 1M for `openai/gpt-5.6-sol`) collapsed to the 200k default, and models with `capabilities.reasoning: false` advertised reasoning controls with a default level.

Now mapped: `capabilities.reasoning`, `capabilities.vision`, `context_length`, `max_output_tokens` (`profiles.py`). Unknown limits stay unknown. Added a discovery-to-Codex-catalog projection test.

### 2. Reasoning intent is preserved instead of forced to `high`
The previous `NamedEffortReasoning(enabled_value="high")` sent `high` for every ON request, ignoring client budgets and adaptive intent. New `XKiroReasoning` encoder:

| Client intent | Wire format |
|---|---|
| OFF | `reasoning_effort: "none"` |
| DEFAULT (no opinion) | field omitted entirely |
| named effort | `reasoning_effort` verbatim (all six values) |
| budget (e.g. `budget_tokens: 4096`) | `extra_body.thinking {"type": "enabled", "budget_tokens": N}` (lossless, no lossy low/medium/high mapping) |
| adaptive / ON without effort | `extra_body.thinking {"type": "adaptive"}` (model applies its own declared default) |

Extras merge with a canonical-field override guard, so a client-supplied `extra_body` cannot clobber the reasoning fields FCC owns.

### 3. Automatic Admin key verification
Previously the Admin UI said "Automatic key verification is unavailable for this provider." xKiro documents `GET /v1/usage` (https://docs.xkiro.com/api/usage/) as key-required, free, no tokens, no rate limit — same `Bearer sk-xt-…` key — so it fits FCC's non-generating credential-check policy. Registered as the probe: `{object: "usage", windows: []}` verifies; `401` (empty JSON error body, verified against the live API) rejects. Confirmed live with a bogus-key request (result: 401).

### Tests
Wire-level regressions via the real SDK serializer: all named efforts, budget boundaries at 6000/24000, DEFAULT key-absence assertion, extras merge without mutating the request, catalog metadata to Codex projection (vision modalities, empty reasoning levels for non-reasoning models), and probe accept/reject cases per docs.xkiro.com. `tests/providers/test_xkiro.py`: 32 passed; credential validation suite: 75 passed; contracts/config/api: 1240 passed; ruff + ty clean.

### Version
6.3.0 to 6.3.1 (`uv.lock` version line only, no dependency changes).

